PR #2, opened 2022-05-14
feat: implement label and enabled in macOS TextMenuItem

pewsheen:
Implement todos (label, enabled) in macOS TextMenuItem

### DIFF
--- a/src/platform_impl/macos/menu_item.rs
+++ b/src/platform_impl/macos/menu_item.rs
@@ -10,7 +10,7 @@ use objc::{
     runtime::{Class, Object, Sel},
     sel, sel_impl,
 };
-use std::ffi::CStr;
+use std::slice;
 use std::sync::Once;
 use std::{
     collections::hash_map::DefaultHasher,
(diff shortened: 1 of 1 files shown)


src/platform_impl/macos/menu_item.rs
```diff
@@ -73,19 +74,36 @@ impl TextMenuItem {
     }
 
     pub fn label(&self) -> String {
-        todo!()
+        unsafe {
+            let title = msg_send![self.ns_menu_item, title];
+            std::str::from_utf8(CStr::from_ptr(title).to_bytes())
+                .unwrap()
+                .to_string()
```
amrbashir, line 81: maybe using `*_lossy` is better to avoid a potential panic?
```suggestion
         std::str::from_utf8_lossy(CStr::from_ptr(title).to_bytes())
                .to_string()
```

amrbashir (changes requested):
This is great, thanks

src/platform_impl/macos/menu_item.rs
pewsheen: Thanks for the reminder, using `*_lossy` will be better.

I just found a proper way to get UTF8String and its length from NSString it self, I'll push another commit later.

amrbashir (approved):

PR #4, opened 2022-06-04
Add `selected` state to `TextMenuItem`

pewsheen:
Added selected methods to the `TextMenuItem` interface and  modified some examples.

I've only implemented it in macOS , so I leave todo!() in windows and unix's impl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,12 @@ Create the root menu and add submenus and men items.
 let mut menu = Menu::new();
 
 let file_menu = menu.add_submenu("File", true);
-let open_item = file_menu.add_text_item("Open", true);
-let save_item = file_menu.add_text_item("Save", true);
+let open_item = file_menu.add_text_item("Open", true, false);
+let save_item = file_menu.add_text_item("Save", true, false);
 
 let edit_menu = menu.add_submenu("Edit", true);
-let copy_item = file_menu.add_text_item("Copy", true);
-let cut_item = file_menu.add_text_item("Cut", true);
+let copy_item = file_menu.add_text_item("Copy", true, false);
+let cut_item = file_menu.add_text_item("Cut", true, false);
 
 #[cfg(target_os = "windows")]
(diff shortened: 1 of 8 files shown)


amrbashir:
Thanks for helping out with this crate.

I don't think this is the right approach. I choose `TextMenuItem` for the default menu item name because I wanted in the future to add `SelectMenuItem`(or `CheckMenuItem`), `ImageMenuItem` and `RadioMenuItem`. This will also comply better with gtk because it doesn't allow to select `gtk::MenuItem` and have another type for that which is [`gtk::CheckMenuItem`](https://docs.gtk.org/gtk3/class.CheckMenuItem.html). cc @wusyong

pewsheen:
> Thanks for helping out with this crate.
> 
> I don't think this is the right approach. I choose `TextMenuItem` for the default menu item name because I wanted in the future to add `SelectMenuItem`(or `CheckMenuItem`), `ImageMenuItem` and `RadioMenuItem`. This will also comply better with gtk because it doesn't allow to select `gtk::MenuItem` and have another type for that which is [`gtk::CheckMenuItem`](https://docs.gtk.org/gtk3/class.CheckMenuItem.html). cc @wusyong

Oh, ok.

This was what I thought if we keep adding things to that interface.

wusyong:
> Thanks for helping out with this crate.
> 
> I don't think this is the right approach. I choose `TextMenuItem` for the default menu item name because I wanted in the future to add `SelectMenuItem`(or `CheckMenuItem`), `ImageMenuItem` and `RadioMenuItem`. This will also comply better with gtk because it doesn't allow to select `gtk::MenuItem` and have another type for that which is [`gtk::CheckMenuItem`](https://docs.gtk.org/gtk3/class.CheckMenuItem.html). cc @wusyong

Is it the same for windows? If so, then let's go for that. NSMenu seems more flexible than others.

amrbashir:
Windows have the most flexible menu system of all 3 platforms so far, usually it is `gtk` that has limitations.

PR #12, opened 2022-06-11
feat(macos): add accelerator

pewsheen:
* implement macOS accelerator
* add `add_native_item` in `Submenu` to make it compiled

### DIFF
--- a/src/platform_impl/macos/accelerator.rs
+++ b/src/platform_impl/macos/accelerator.rs
@@ -1,0 +1,112 @@
+use cocoa::appkit::NSEventModifierFlags;
+
+/// Mnemonic is deprecated since macOS 10
+pub fn remove_mnemonic(string: impl AsRef<str>) -> String {
+    string
+        .as_ref()
+        .replace("&&", "[~~]")
+        .replace("&", "")
+        .replace("[~~]", "&")
+}
+
+/// Returns a tuple of (Key, Modifier)
+pub fn parse_accelerator(accelerator: impl AsRef<str>) -> (String, NSEventModifierFlags) {
+    let accelerator = accelerator.as_ref();
+    let mut s = accelerator.split("+");
+    let count = s.clone().count();
+    let (mod1, mod2, key) = {
+        if count == 2 {
+            (s.next().unwrap(), None, s.next().unwrap())
+        } else if count == 3 {
+            (
+                s.next().unwrap(),
+                Some(s.next().unwrap()),
+                s.next().unwrap(),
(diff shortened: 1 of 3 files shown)


src/platform_impl/macos/accelerator.rs
```diff
@@ -0,0 +1,112 @@
+use cocoa::appkit::NSEventModifierFlags;
+
+/// Mnemonic is deprecated since macOS 10
+pub fn remove_mnemonic(string: impl AsRef<str>) -> String {
+    string
+        .as_ref()
+        .replace("&&", "[~~]")
+        .replace("&", "")
+        .replace("[~~]", "&")
+}
```
amrbashir, line 10:
```suggestion
/// Mnemonic is deprecated since macOS 10
pub fn remove_mnemonic(string: impl AsRef<str>) -> String {
    string
        .as_ref()
        .replace("&", "")
}
```

src/platform_impl/macos/menu_item.rs
```diff
@@ -54,7 +56,7 @@ impl TextMenuItem {
 
     pub fn set_label(&mut self, label: impl AsRef<str>) {
         unsafe {
-            let title = NSString::alloc(nil).init_str(label.as_ref());
+            let title = NSString::alloc(nil).init_str(&remove_mnemonic(label));
```
amrbashir, line 59: Since we are removing the mnemonic when setting the `label`, we should add it back when returning the label in `TextMenuItem::label`

amrbashir (changes requested):

src/platform_impl/macos/accelerator.rs
pewsheen: Hmm, what if we want to have "&" in the label like "Save & Close"

src/platform_impl/macos/menu_item.rs
pewsheen: Seems It should be store somewhere in the `TextMenuItem` struct? Since ampersands were removed from the `NSMenuItem`, we can't retrieve from it.

Something like this
```rust
pub struct TextMenuItem {
    pub(crate) id: u64,
    pub(crate) ns_menu_item: id,
    label: String,
}
```

amrbashir: yeah that sounds fine, although you want to wrap the label in `Rc` because `TextMenuItem` derives `Clone` and we want the clone to also point to the same label.

amrbashir:
Thank you!

PR #17, opened 2022-07-19
Add accelerator module

wusyong:


src/accelerator.rs
```diff
@@ -0,0 +1,467 @@
...
+            SysMods::None => RawMods::None,
+            SysMods::Shift => RawMods::Shift,
+            SysMods::Cmd => RawMods::Ctrl,
+            SysMods::AltCmd => RawMods::AltCtrl,
+            SysMods::CmdShift => RawMods::CtrlShift,
+            SysMods::AltCmdShift => RawMods::AltCtrlShift,
+        }
+    }
+}
+
```
amrbashir, line 250:
```suggestion
```

src/accelerator.rs
```diff
@@ -0,0 +1,467 @@
...
+    pub fn new(accelerator_string: &str) -> AcceleratorId {
+        AcceleratorId(hash_string_to_u16(accelerator_string))
+    }
+
+    /// Whenever this menu is empty.
+    pub fn is_empty(self) -> bool {
+        Self::EMPTY == self
+    }
+}
+
```
amrbashir, line 287: Not needed
```suggestion
```

src/accelerator.rs
```diff
@@ -0,0 +1,467 @@
...
+    title.to_uppercase().hash(&mut s);
+    s.finish() as u16
+}
+
+fn hash_accelerator_to_u16(hotkey: Accelerator) -> u16 {
+    let mut s = DefaultHasher::new();
+    hotkey.hash(&mut s);
+    s.finish() as u16
+}
+
```
amrbashir, line 378: Not needed, just use `utill::Counter` to generate ids
```suggestion
```

src/accelerator.rs
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /// Returns an identifier unique to the accelerator.
+    pub fn id(self) -> AcceleratorId {
+        if let Some(id) = self.id {
+            return id;
+        }
+
+        AcceleratorId(hash_accelerator_to_u16(self))
+    }
```
amrbashir, line 82: Not needed. Accelerator is kind of a metadata, they serve nothing on their own. They don't need ids.

amrbashir (changes requested):

wusyong:
There are some doc tests failed outside of this PR.
But I think we should address it.
I'll fix it tomorrow.

examples/tao.rs
```diff
@@ -19,7 +23,11 @@ fn main() {
 
     let mut file_menu = menu_bar.add_submenu("&File", true);
     let mut open_item = file_menu.add_item("&Open", true, None);
-    let mut save_item = file_menu.add_item("&Save", true, Some("CommandOrCtrl+S"));
+    let mut save_item = file_menu.add_item(
+        "&Save",
+        true,
+        Some(Accelerator::new(Mods::CtrlMeta, Code::KeyS)),
```
amrbashir, line 29:
```suggestion
        Some(Accelerator::new(Mods::Ctrl, Code::KeyS)),
```

examples/winit.rs
```diff
@@ -37,7 +41,11 @@ fn main() {
 
     let mut file_menu = menu_bar.add_submenu("&File", true);
     let mut open_item = file_menu.add_item("&Open", true, None);
-    let mut save_item = file_menu.add_item("&Save", true, Some("CommandOrCtrl+S"));
+    let mut save_item = file_menu.add_item(
+        "&Save",
+        true,
+        Some(Accelerator::new(Mods::CtrlMeta, Code::KeyS)),
```
amrbashir, line 47:
```suggestion
        Some(Accelerator::new(Mods::Ctrl, Code::KeyS)),
```

src/accelerator.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    Ok(Accelerator { key, mods })
+}
+
+#[test]
+fn test_parse_accelerator() {
+    assert_eq!(
+        parse_accelerator("CTRL+KeyX").unwrap(),
```
amrbashir, line 213: This notation of `KeyX` will definitely break tauri so temporarily until tauri v2, we are gonna need to make a compatibility layer but not sure whether it should be in muda or tauri.
cc @lucasfernog

src/accelerator.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Represents the active modifier keys.
+///
+/// This is intended to be clearer than [`Modifiers`], when describing accelerators.
+///
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Hash)]
+pub enum Mods {
+    None,
+    Alt,
+    Ctrl,
```
amrbashir, line 83: Missing variants for macos
```suggestion
    Ctrl,
	Command,
    CommandOrCtrl,
```

amrbashir (changes requested):

src/accelerator.rs
wusyong: How about we just fork the types in `keyboard_types` then. I feel like it doesn't need to stick to origin crate so hard.
We only need its `Code` and `Modifiers` types.

amrbashir: No need to fork, we just need to have an internal function that takes the `Code` and converts it to a string that is compatible with current tao implementation. We only need to keep it around until tauri v2

amrbashir (approved):

PR #18, opened 2022-10-06
refactor: rewrite

### LINKED ISSUE #5: Tracking issue for missing functionality in `muda`
### Necessary:
- [X] `init_for_window`
- [X] `hide_for_window` - #8
- [X] `show_for_window` - #8
- [X] `remove_for_window` - #8
- [X] `Submenu` 
- [X] `MenuItem`
- [X] Accelerators
  - [X] Windows - #6
  - [X] Linux - #6
  - [X] macOS  - #12
- [x] `NativeMenuItem`
  - [X] Windows - #9 
  - [X] Linux - #9
  - [x] macOS - #14
- [x] `CheckMenuItem`
  - [x] Windows - #13
  - [x] Linux - #13
  - [x] macOS - #14 
- [ ] ~~Localized Labels~~
- [ ] Change native items labels
- [X] `Accelerator::from_str()` compatibility with current tao implementation - #17
  
### Extras
- [ ] `Menu::showContextMenu(window, x, y)`
  - [ ] Windows - Possible, use `TrackPopupMenu` from win32 apis
  - [ ] Linux - Possible, use `gtk_menu_popup_at_rect` from gtk
  - [ ] macOS - Possible, use `popUpMenuPositioningItem` from appkit
- [ ] `ImageMenuItem`
  - [ ] Windows
  - [ ] Linux
  - [ ] macOS
- [ ] `RadioMenuItem`
  - [ ] Windows
  - [ ] Linux
  - [ ] macOS

amrbashir:
This rewrite main goal is to allow modifying the predefined (native) menu items and allow creating the menu item separately from a menu.

- [x] gtk
- [X] Windows
- [x] macOS
- [X] Error handling
- [x] Documentation
- [ ] `show_context_menu_for_ns_view` should take an `x` and `y`.


PerBothner:
As requested [here](https://github.com/tauri-apps/wry/issues/677#issuecomment-1275419120) some niceties of [jsMenus](https://github.com/PerBothner/jsMenus) and other wishlist items. Note I haven't really looked at Muda or Wry's menu-creating API - currently DomTerm can use the Electron or Qt menus or jsMenus.

* A way to express multi-key shortcuts.  These are common in Emacs.  They are also used in the Atom editor - check out the View->Panes submenu.  Notice that multi-key shortcuts appear in the menu-item in square brackets as part of the label text like `Split Left [Ctrl+K Left]`, while single-key shortcuts appear right-aligned without the square brackets. This is presumably due to Electron limitations. Note I'm not asking for the menu API to be able to process multikey shortcuts - only that they can be displayed in a non-ugly manner.
* In DomTerm when using a custom titlebar, the menubar is included in the titlebar (except on macOS with the Electron or Qt front-ends), which saves space.  This isn't supported by most toolkits, though it could be simulated with a row of buttons that bring up a popup (context) menu.
* A nice API to modify menus or generate them on the fly (such as a list of sub-windows).  (Electron doesn't seem to have this, except by re-creating the menu(s) from scratch: You can modify or add menuitems, but there doesn't seem to be a way to replace or delete them.)
* The Electron `Menu.buildFromTemplate` is nice as it allows me to build the menu in the "renderer" process and send it to the main process by ipc.
* Potentially it would be nice to include rich text in menu labels, based on HTML or a subset thereof.  (jsMenus doesn't currently implement that, but it would be easy to add, as it's all implemented by HTML elements and JavaScript.)

amrbashir:
> * A way to express multi-key shortcuts.  These are common in Emacs.  They are also used in the Atom editor - check out the View->Panes submenu.  Notice that multi-key shortcuts appear in the menu-item in square brackets as part of the label text like `Split Left [Ctrl+K Left]`, while single-key shortcuts appear right-aligned without the square brackets. This is presumably due to Electron limitations. Note I'm not asking for the menu API to be able to process multikey shortcuts - only that they can be displayed in a non-ugly manner.

I maybe able to do that but I won't include it in this PR

> * In DomTerm when using a custom titlebar, the menubar is included in the titlebar (except on macOS with the Electron or Qt front-ends), which saves space.  This isn't supported by most toolkits, though it could be simulated with a row of buttons that bring up a popup (context) menu.

Not sure if I understand this correctly, do you want to display a native menubar when the native titlebar is hiddden?

> * A nice API to modify menus or generate them on the fly (such as a list of sub-windows).  (Electron doesn't seem to have this, except by re-creating the menu(s) from scratch: You can modify or add menuitems, but there doesn't seem to be a way to replace or delete them.)

This PR implements all necessary menu modification APIs, `append`, `prepend`, `insert`, `append_items`, `prepend_items`, `insert_items` and `remove`

> * The Electron `Menu.buildFromTemplate` is nice as it allows me to build the menu in the "renderer" process and send it to the main process by ipc.

This is out of Tao/Wry/Muda scope, but it is possible that Tauri will provide a JS api for it in the future.

> * Potentially it would be nice to include rich text in menu labels, based on HTML or a subset thereof.  (jsMenus doesn't currently implement that, but it would be easy to add, as it's all implemented by HTML elements and JavaScript.)

This is not possible unfortunately, as these are native menus and rendering is done by the OS.

PerBothner:
_"Not sure if I understand this correctly, do you want to display a native menubar when the native titlebar is hiddden?"_

I'd like the ability to embed the menubar into a custom titlebar, as shown below. When space is tight (or depending on user/application preference) the menubar may be replaced by a hamburger menu.

Screenshot from 2022-10-12 11-48-29

amrbashir:
I don't think that is possible, because the custom titlebar would be in the webview while the menu is in the native window.

For this situation you will have to fallback to a menu implementation in HTML/JS, similar to how VSCode has a native menu when using native title bar and a custom HTML/JS menu when using custom title bar.

PerBothner:
_"This is not possible unfortunately, as these are native menus and rendering is done by the OS."_

I understand - though it is tempting to stick to jsMenus as opposed to native menus.  For the record, I see these as the advantages of native menus:

* On macOS, use the system menubar. I think this is the main advantage of native windows, at least for DomTerm.
* Menus can stick outside the window boundary - though that can be worked around by using a popup window.
* Native menu styling - though I'm not sure there is a such a thing (except perhaps on macOS), since there are so many toolkits, that create very different-looking menus.
* Avoid need for a webview.

PerBothner:
"_I don't think that is possible, because the custom titlebar would be in the webview while the menu is in the native window."_

The custom titlebar could have a transparent notch allowing the native menubar to be visible. Or (at some point in the future) use something like the new experimental [Windows Controls Overlay API](https://developer.mozilla.org/en-US/docs/Web/API/Window_Controls_Overlay_API).

_"For this situation you will have to fallback to a menu implementation in HTML/JS, similar to how VSCode has a native menu when using native title bar and a custom HTML/JS menu when using custom title bar."_

Yes, that is more or less what I do. There is also a hybrid option: custom buttons for the menubar itself, but when you click on one you get a native popup menu.

PerBothner:
_"you still need to a way pass the click to the window rather than the webview"_

That is why I fiddle with z-order when a menu (using jsMenus) is shown. Specifically, the "master" webview is brought to the front so it can get mouse and keyboard events.

PerBothner:
I agree for Windows or Linux. On macOS I think it makes sense to use the system menubar even if using a custom titlebar - and in tha case we don't have to worry about embedding the menubar into the titlebar

One advantage of an API like Electron's `buildFromTemplate` is one can use shared logic for building menus, and use it for both native menus and jsMenus. I currently do that for Electron menus. I hope to do something similar for both Qt and in the future Muda: A utility function to translate a template into a menu structure. Easier for me with Qt because I'm more fluent with C++ than Rust.

amrbashir:
Sounds reasonable, and I already implemented template support using `Menu::append_items` in this PR

src/lib.rs
```diff
@@ -58,59 +68,186 @@
...
+    /// ## Panics
     ///
-    /// - **Windows / Linux:** The menu label can contain `&` to indicate which letter should get a generated accelerator.
-    /// For example, using `&File` for the File menu would result in the label gets an underline under the `F`,
-    /// and the `&` character is not displayed on menu label.
-    /// Then the menu can be activated by press `Alt+F`.
-    pub fn add_submenu<S: AsRef<str>>(&mut self, label: S, enabled: bool) -> Submenu {
-        Submenu(self.0.add_submenu(label, enabled))
+    /// - If `item` has already been removed
+    /// - If `item` wasn't previously [append](Menu::append)ed to this menu
```
caesar, line 243: Wouldn't a graceful failure (such as taking no action) be better? Or maybe returning a `Result` if we want to let the caller know there was a problem, though I don't think that would be necessary here.

src/lib.rs
```diff
@@ -58,59 +68,186 @@
...
+    /// Add menu items to the beginning of this menu.
+    /// It calls [`Menu::prepend`] on the first element and
+    /// passes the rest to [`Menu::insert_items`] with position of `1`.
+    ///
+    /// ## Platform-spcific:
+    ///
+    /// - **macOS:** Only [`Submenu`] can be added to the menu
+    pub fn prepend_items(&self, items: &[&dyn MenuItemExt]) {
+        self.prepend(items[0]);
+        self.insert_items(&items[1..], 1);
```
caesar, line 215: Can't we do `insert_items` with a position of `0`, instead of special-casing the first item?

src/lib.rs
amrbashir: I will add error handling in another PR, the library shouldn't panic at most cases.

src/lib.rs
amrbashir: yeah you're right

amrbashir: yeah you're right

PR #19, opened 2022-10-18
macOS implementation

caesar:
***This is far from complete***, but I thought I'd open a draft pull request so you can see where I'm going with it and let me know any major comments before I do too much more.

There is a lot that will have to be changed, and custom menu item actions don't do anything yet, but most other stuff is working.

Note that I've also added some extra predefined menu items but I haven't implemented them for platforms other than macOS.


src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+        };
+
+        unsafe {
+            match op {
+                AddOp::Append => self.ns_menu.addItem_(ns_menu_item),
+                AddOp::Insert(position) => {
+                    msg_send![self.ns_menu, insertItem: ns_menu_item atIndex: position as NSInteger]
+                }
+            }
+        }
```
amrbashir, line 154: `Menu` should also contain a `children` field and you should add to it after this call

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+            for (_, &ns_item) in &self.ns_menuitems {
+                let () = msg_send![ns_item, setState: if checked { 1_isize } else { 0_isize }];
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct Menu {
+    pub ns_menu: id,
```
amrbashir, line 102: why is `ns_menu` public?

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+        unsafe {
+            for (_, &ns_item) in &self.ns_menuitems {
+                let () = msg_send![ns_item, setState: if checked { 1_isize } else { 0_isize }];
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct Menu {
```
amrbashir, line 101: should also hold a `children` field of its children.

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+                menuitem.0.ns_item_for_menu(self)
+            }
+            MenuEntryType::Predefined => {
+                let menuitem = item.as_any().downcast_ref::<crate::PredefinedMenuItem>().unwrap();
+                menuitem.0.ns_item_for_menu(self)
+            }
+        };
+        unsafe {
+            let () = msg_send![self.ns_menu, removeItem: ns_menu_item];
+        }
```
amrbashir, line 178: `Menu` should also contain a `children` field and you should remove from it after this call

src/lib.rs
```diff
@@ -255,6 +259,11 @@ impl Menu {
         self.0.show_context_menu_for_hwnd(hwnd, x, y)
     }
 
+    #[cfg(target_os = "macos")]
+    pub fn show_context_menu_for_nsview(&self, view: cocoa::base::id) {
```
amrbashir, line 263: This method should also take an `x`, `y` position as to where to show the menu, does macOS not have this?

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct PredefinedMenuItem(Rc<RefCell<MenuChild>>);
+
+impl PredefinedMenuItem {
+    pub fn new(item_type: PredfinedMenuItemType, text: Option<String>) -> Self {
+        let text = text
+            .unwrap_or_else(|| item_type.text().to_string())
+            .replace("&", "");
```
amrbashir, line 348: this will remove all `&`, we should only remove single `&` and keep `&&` and convert it to a single `&`

The rule is, `&` before a char defines an auto mnemonic for this menu item, if you want to display `&` as is, you should escape it and use `&&`

amrbashir (commented):
Overall very good, left some comments along with a few nit-picks that you can ignore for now.

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+/// Be careful when cloning this item and treat it as read-only
+#[derive(Debug, Default)]
+#[allow(dead_code)]
+struct MenuChild {
+    // shared fields between submenus and menu items
+    type_: MenuEntryType,
+    text: String,
+    enabled: bool,
+
+    ns_menuitems: HashMap<Menu, id>,
```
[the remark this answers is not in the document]
caesar: Yes, adding the same item to multiple menus causes exceptions unfortunately.

One thing that's not solved by the current implementation is adding the same item *to the same menu* multiple times. Is that something we care about…? What did you do about that on GTK? (Guess I can just have a look…)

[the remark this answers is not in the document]
caesar: I should also note that I'm going to have to change `HashMap<Menu, id>` to `HashMap<u32, id>` and index by the parent menu's `id` field instead, because I need to add more fields to `Menu` and it will become hard to derive `Eq` which is needed for it to be a `HashMap` key.

src/lib.rs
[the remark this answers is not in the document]
caesar: Happy to change that, but what's the advantage?

For what it's worth in my testing it's not actually really required to pass an NSView at all, `nil` works fine…
I guess there must be cases where it's needed though.

src/platform_impl/macos/mod.rs
caesar: Yup, that's next on my list of things to implement, I'll also need it for `items()`. I just haven't done it yet because of the `HashMap` change I mentioned above but I'll make that change later today.

src/platform_impl/macos/mod.rs
caesar: Good point, I don't think it needs to be.

src/lib.rs
caesar: It's possible to open a context menu at an `x` and `y`, but it's not really needed and it's also quite a bit more complicated because coordinate system conversions will be needed that I haven't quite figured out.

At the moment I'm just opening the menu at the location of the "current event", which should always be where the click was that triggered the menu to be shown.

On the other hand I can see that opening a menu at specified coordinates would be useful for other purposes than just context menus (eg dropdown menus) so I'll come back to that and try and figure out the coordinate system transforms.

src/platform_impl/macos/mod.rs
caesar: Ah yeah I did wonder about that. I guess some regex will be needed 😑

caesar: Do the 'mnemonics' work on Windows btw? I couldn't find any evidence of them when I ran the example on Win 11.

src/lib.rs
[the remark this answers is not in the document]
amrbashir: I think we need to use https://developer.apple.com/documentation/appkit/nsmenu/1518212-popupmenupositioningitem?language=objc

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,526 @@
...
+                let ns_submenu: *mut Object = msg_send![ns_item, submenu];
+                if ns_submenu != nil {
+                    let () = msg_send![ns_submenu, setTitle: title];
+                }
+            }
+        }
+    }
+
+    pub fn is_enabled(&self) -> bool {
+        self.enabled
```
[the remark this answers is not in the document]
amrbashir: We need to make sure all instances are synced up. Check menu items for example, when triggered in one menu using a click, doesn't mean it is checked in other menus so we loop through the instances and sync it

src/platform_impl/macos/mod.rs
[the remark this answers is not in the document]
amrbashir: On gtk, each item has a hashmap whefe each key is the parent menu id and the value is a vector of gtk items, in macOS case, it will be ns_menuitem

src/lib.rs
amrbashir: yeah we need to support x and y, and maybe in the future add a variant that opens under the cursor position

src/platform_impl/macos/mod.rs
amrbashir: Yeah they do, press Alt and you should see a character getting underline

amrbashir: > Ah yeah I did wonder about that. I guess some regex will be needed 😑

No need for regex, take a look at a dirty way to doing it in gtk/accelerator.ra

src/lib.rs
caesar: Yeah ok, I'll have another look at that.
Do you have any ideas about the coordinate transforms we need? There seems to be a scaling issue (I guess logical vs physical pixels?) as well as flipping the `y` coordinate.

amrbashir: no need for conversions, just assume that x and y are logical and relative to the view top-left corner

caesar: Just providing x and y as coordinates to https://developer.apple.com/documentation/appkit/nsmenu/1518212-popupmenupositioningitem?language=objc doesn't work at all, there's at least a scaling issue for both axes and a sign flip for the y axis (AppKit uses coordinates relative to bottom left 🙄), and possibly also an offset of some sort. 

I'll play with it again later but it was driving me crazy trying to figure it out when I looked at it a few days ago...

src/platform_impl/macos/mod.rs
caesar: > Yeah they do, press Alt and you should see a character getting underline

Just tried again and I'm definitely not seeing that in my Windows 11 VM, all that happens is when I release Alt the menu closes.

src/lib.rs
amrbashir: sign flip should be fine, scaling should also be possible, for instance Windows has `ClientToScreen` api

src/platform_impl/macos/mod.rs
amrbashir: I will double check that but mnemonics are not supported on macOS iirc, so you only need to remove single `&` and convert `&&` to `&`

src/lib.rs
amrbashir: maybe this could help https://developer.apple.com/documentation/uikit/uiview/1622498-convertrect

src/platform_impl/macos/mod.rs
amrbashir: There is definitely something weird with the menubar mnemonics but you can press `Alt+F` to open the file submenu and you should see the menu items have the mnemonics displayed correctly. I will see if I can fix the menubar.

src/lib.rs
caesar: Yeah I did try a few similar methods. I think I'll have to look into how Tao / Winit / whatever crate extracts the coords from the original NSEvent, and do the reverse.

I'll get to it later or tomorrow; for now I'm refactoring a lot of other stuff to make it easier to implement the remaining features.

src/platform_impl/macos/mod.rs
amrbashir: The menubar issue is something that needs to be fixed in tao and winit eventloop. I will open a PR to fix it. Thanks for catching this.

src/lib.rs
amrbashir: You can leave the context menu for some other time, maybe in a different PR if needed

src/platform_impl/macos/mod.rs
caesar: Ah, that's great that you've tracked it down!

> mnemonics are not supported on macOS iirc, so you only need to remove single `&` and convert `&&` to `&`

Yup that's right, macOS doesn't support them anymore.

src/lib.rs
caesar: Yes okay, if I don't find an easy way I'll do that. The current method works for context menus anyway, so it's only for popup menus of other kinds that it will be needed.

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,674 @@
...
+pub(crate) struct PredefinedMenuItem(Rc<RefCell<MenuChild>>);
+
+impl PredefinedMenuItem {
+    pub fn new(item_type: PredfinedMenuItemType, text: Option<String>) -> Self {
+        let text = strip_mnemonic(text.unwrap_or_else(|| match item_type {
+            PredfinedMenuItemType::About(_) => format!("About {}", unsafe { app_name_string() }.unwrap_or_default()).trim().to_string(),
+            PredfinedMenuItemType::Hide => format!("Hide {}", unsafe { app_name_string() }.unwrap_or_default()).trim().to_string(),
+            PredfinedMenuItemType::Quit => format!("Quit {}", unsafe { app_name_string() }.unwrap_or_default()).trim().to_string(),
+            _ => item_type.text().to_string(),
+        }));
```
caesar, line 423: @amrbashir I'm not sure about the approach I've taken here because i18n will be hard / impossible.

I think a better approach would be to add placeholders in the output of [`PredefinedMenuItem::text()`](https://github.com/amrbashir/muda/pull/19/files#diff-ade96ad6ce10818697881295338193a079d0df5ebf874621467084a423951b9bR39) – maybe something like `[APP_NAME]` could be replaced by the actual app name (only on macOS… at least for now?)

That way localized text provided by the user could contain the same placeholders.

What do you think?

caesar (commented):
Getting there… still a few things to fix. I'll see what I can get done tomorrow (Monday); after that I'm probably going to be unavailable for a while.

Please see my comment / question below @amrbashir.

src/platform_impl/macos/mod.rs
amrbashir: That's not necessary. If they want to do i18n for a predefined item, they need to also provide the app name if they want to display it.

caesar: Ok, I'll leave it as is then, at least for now.

amrbashir:
Hope you don't mind, I just added error handling in the rewrite branch and merged it here as well.

src/lib.rs
```diff
@@ -180,12 +198,19 @@ pub trait ContextMenu {
...
     /// Get the underlying gtk menu reserved for context menus.
     #[cfg(target_os = "linux")]
     fn gtk_context_menu(&self) -> gtk::Menu;
+
+    /// Shows this menu as a context menu for the specified `NSView`.
+    ///
+    /// The menu will be shown at the coordinates of the current event
+    /// (the click which triggered the menu to be shown).
+    #[cfg(target_os = "macos")]
+    fn show_context_menu_for_nsview(&self, view: cocoa::base::id);
```
amrbashir, line 211: Still want to see if it is possible to use an x & y coordinate instead of relying on the cursor position, @wusyong do you have some info about it? possibly `popUpMenuPositioningItem `

src/platform_impl/macos/mod.rs
```diff
@@ -1 +1,855 @@
...
+                let item_type = &item.borrow().type_.clone();
+                let ns_item = match item_type {
+                    MenuItemType::Submenu => Submenu(item.clone()).make_ns_item_for_menu(menu_id),
+                    MenuItemType::Normal => MenuItem(item.clone()).make_ns_item_for_menu(menu_id),
+                    MenuItemType::Predefined => {
+                        PredefinedMenuItem(item.clone()).make_ns_item_for_menu(menu_id)
+                    }
+                    MenuItemType::Check => {
+                        CheckMenuItem(item.clone()).make_ns_item_for_menu(menu_id)
+                    }
```
amrbashir: `menu_id` here is the parent menu, shouldn't you create a new `menu_id` for this newly created submenu and use that new id for the new children?

src/lib.rs
caesar: Yes, that would be great. Unfortunately I don't currently have the time to do any further research on the coordinate issues mentioned above, but if anyone else can find a way that would be great, or perhaps it can be added later.

amrbashir:
Thank you @caesar

PR #24, opened 2022-11-30
refactor: rewrite gtk backend from scratch

amrbashir:


FabianLars (approved):
Works fine as far as i can see on wsl2+wslg+win10 and fedora37+KDE 👍 

There's only the warning we talked about on Discord which seemingly has no effect on functionality.

amrbashir:
thanks @FabianLars

PR #66, opened 2023-06-19
support `AboutMetadata` on macos

mxsdev:
Passes `AboutMetadata` to the About predefined menu type. Data is passed through to `orderFrontStandardAboutPanelWithOptions`.

I had to introduce a few new fields to support e.g. `NSAboutPanelOptionApplicationIcon`.

### DIFF
--- a/src/platform_impl/macos/mod.rs
+++ b/src/platform_impl/macos/mod.rs
@@ -653,8 +653,8 @@ impl PredefinedMenuItem {
             _ => create_ns_menu_item(&child.text, item_type.selector(), &child.accelerator),
         };
 
-        unsafe {
-            if let PredfinedMenuItemType::About(_) = child.predefined_item_type {
+        if let PredfinedMenuItemType::About(_) = child.predefined_item_type {
+            unsafe {
                 let _: () = msg_send![ns_menu_item, setTarget: ns_menu_item];
                 let _: () = msg_send![ns_menu_item, setTag:child.id()];
 
--- a/src/platform_impl/macos/mod.rs
+++ b/src/platform_impl/macos/mod.rs
@@ -1009,61 +1009,61 @@ extern "C" fn fire_menu_item_click(this: &Object, _: Sel, _item: id) {
         let ptr: usize = *this.get_ivar(BLOCK_PTR);
         let item = ptr as *mut &mut MenuChild;
 
-        match &(*item).predefined_item_type {
-            PredfinedMenuItemType::About(Some(about_meta)) => {
-                unsafe fn mkstr(s: &str) -> id {
-                    NSString::alloc(nil).init_str(s)
+        if let PredfinedMenuItemType::About(about_meta) = &(*item).predefined_item_type {
+            match about_meta {
+                Some(about_meta) => {
+                    unsafe fn mkstr(s: &str) -> id {
+                        NSString::alloc(nil).init_str(s)
+                    }
+
+                    let mut keys: Vec<id> = Default::default();
+                    let mut objects: Vec<id> = Default::default();
+
+                    if let Some(name) = &about_meta.name {
+                        keys.push(NSAboutPanelOptionApplicationName);
+                        objects.push(mkstr(&name));
+                    }
+
+                    if let Some(version) = &about_meta.version {
+                        keys.push(NSAboutPanelOptionApplicationVersion);
+                        objects.push(mkstr(&version));
+                    }
+
+                    if let Some(short_version) = &about_meta.short_version {
+                        keys.push(NSAboutPanelOptionVersion);
+                        objects.push(mkstr(&short_version));
+                    }
+
+                    if let Some(copyright) = &about_meta.copyright {
+                        keys.push(NSString::alloc(nil).init_str("Copyright"));
+                        objects.push(mkstr(&copyright));
+                    }
+
+                    if let Some(icon) = &about_meta.icon {
+                        keys.push(NSAboutPanelOptionApplicationIcon);
+                        objects.push(icon.inner.to_nsimage(None));
+                    }
+
+                    if let Some(credits) = &about_meta.credits {
+                        keys.push(NSAboutPanelOptionCredits);
+                        let attributed_str: id = msg_send![class!(NSAttributedString), alloc];
+                        let _: () = msg_send![attributed_str, initWithString: mkstr(&credits)];
+                        objects.push(attributed_str);
+                    }
+
+                    let keys_array = NSArray::arrayWithObjects(nil, &keys);
+                    let objs_array = NSArray::arrayWithObjects(nil, &objects);
+
+                    let dict =
+                        NSDictionary::dictionaryWithObjects_forKeys_(nil, objs_array, keys_array);
+
+                    let _: () = msg_send![NSApp(), orderFrontStandardAboutPanelWithOptions: dict];
                 }
 
-                let mut keys: Vec<id> = Default::default();
-                let mut objects: Vec<id> = Default::default();
-
-                if let Some(name) = &about_meta.name {
-                    keys.push(NSAboutPanelOptionApplicationName);
-                    objects.push(mkstr(&name));
+                None => {
+                    let _: () = msg_send![NSApp(), orderFrontStandardAboutPanel: this];
                 }
-
-                if let Some(version) = &about_meta.version {
-                    keys.push(NSAboutPanelOptionApplicationVersion);
-                    objects.push(mkstr(&version));
-                }
-
-                if let Some(short_version) = &about_meta.short_version {
-                    keys.push(NSAboutPanelOptionVersion);
-                    objects.push(mkstr(&short_version));
-                }
-
-                if let Some(copyright) = &about_meta.copyright {
-                    keys.push(NSString::alloc(nil).init_str("Copyright"));
-                    objects.push(mkstr(&copyright));
-                }
-
-                if let Some(icon) = &about_meta.icon {
-                    keys.push(NSAboutPanelOptionApplicationIcon);
-                    objects.push(icon.inner.to_nsimage(None));
-                }
-
-                if let Some(credits) = &about_meta.credits {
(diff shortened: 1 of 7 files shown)


src/platform_impl/macos/mod.rs
```diff
@@ -988,6 +1009,63 @@ extern "C" fn fire_menu_item_click(this: &Object, _: Sel, _item: id) {
...
+                        objects.push(mkstr(&version));
+                    }
+
+                    if let Some(short_version) = &about_meta.short_version {
+                        keys.push(NSAboutPanelOptionVersion);
+                        objects.push(mkstr(&short_version));
+                    }
+
+                    if let Some(copyright) = &about_meta.copyright {
+                        keys.push(NSString::alloc(nil).init_str("Copyright"));
```
mxsdev: Wasn't sure how to make this a global constant. I wasn't able to do it using `static` since `*mut Object` is not thread-safe.

src/predefined.rs
```diff
@@ -178,31 +178,69 @@ impl PredefinedMenuItem {
...
     /// The application name.
     pub name: Option<String>,
     /// The application version.
     pub version: Option<String>,
+    /// The short version, e.g. "1.0"
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows / Linux:** This is ignored
+    pub short_version: Option<String>,
```
amrbashir: Let's not add this one for the sake of cross-platform support and it doesn't seem to provide much value, like you can just use the `version` field and put whatever you want there.

src/platform_impl/macos/mod.rs
amrbashir: that's fine, you can just make the `Copyright` string a constant and use it
```suggestion
                        keys.push(mkstr(NSAboutPanelOptionCopyright));
```

src/predefined.rs
```diff
@@ -178,31 +178,69 @@ impl PredefinedMenuItem {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** This is ignored
     pub comments: Option<String>,
     /// The copyright of the application.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **macOS:** This is ignored
```
amrbashir: This docs is incorrect, You added an implementation for this

src/predefined.rs
```diff
@@ -178,31 +178,69 @@ impl PredefinedMenuItem {
...
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows / Linux:** This is ignored
+    pub credits: Option<String>,
+    /// The application icon.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows / Linux:** This is ignored
```
amrbashir: This could be implemented on Linux as well, if you want to add that in this PR too, go ahead, otherwise I can add it later

src/platform_impl/macos/mod.rs
amrbashir: I couldn't find an official documentation for this key, how does this render out? can we do the same for other about metadata fields?

amrbashir (changes requested):

src/predefined.rs
```diff
@@ -178,31 +178,69 @@ impl PredefinedMenuItem {
...
+    pub icon: Option<Icon>,
 }
 
-#[derive(PartialEq, Eq, Debug, Clone)]
```
amrbashir, line 205: why was this change made?

src/platform_impl/macos/mod.rs
```diff
@@ -640,11 +649,22 @@ impl PredefinedMenuItem {
...
+                (*ns_menu_item).set_ivar(BLOCK_PTR, ptr as usize);
+            }
+        }
+
         unsafe {
             if !child.enabled {
                 let () = msg_send![ns_menu_item, setEnabled: NO];
             }
-            if child.predefined_item_type == PredfinedMenuItemType::Services {
+            if let PredfinedMenuItemType::Services = child.predefined_item_type {
```
amrbashir, line 671: and subsequently this one

amrbashir (changes requested):

amrbashir:
Thanks for your contribution, this is really awesome to see.

src/predefined.rs
mxsdev: good catch, my bad!

src/predefined.rs
mxsdev: Since `Icon` doesn't implement `PartialEq`.

src/predefined.rs
mxsdev: I see where you're coming from, but it is definitely not uncommon to see this field used across applications. In the interest of not restricting this functionality, maybe it would make sense to put it behind a `#cfg(target_os = "macos")` flag?

image

image

src/platform_impl/macos/mod.rs
mxsdev: [It's a little hidden](https://developer.apple.com/documentation/appkit/nsapplication/1428479-orderfrontstandardaboutpanelwith?language=objc):

> ### Discussion
> In addition to the keys in AboutPanelOptionKey, you may also include the following key in optionsDictionary:
>
> `@"Copyright"`: An NSString object with a line of copyright information. If not specified, this method then looks for the value of NSHumanReadableCopyright in the localized version of the app's Info.plist file. If neither is available, this method leaves the space blank.

src/platform_impl/macos/mod.rs
mxsdev: See other comment

src/predefined.rs
mxsdev: Will do 👍

src/predefined.rs
amrbashir: I think we can extend it to other platforms as well by doing
```rs
fromat!("{}{}", version, short_version.map(|v| format!("({v})")).unwrap_or_default())
```
> Assuming the underlined versions are `short_version`

src/predefined.rs
amrbashir: I see, it is fine then.

src/platform_impl/macos/mod.rs
amrbashir: I was kinda hoping that it would allow providing more keys but it is what it is

src/platform_impl/macos/mod.rs
```diff
@@ -640,11 +653,22 @@ impl PredefinedMenuItem {
...
+            if let PredfinedMenuItemType::About(_) = child.predefined_item_type {
+                let _: () = msg_send![ns_menu_item, setTarget: ns_menu_item];
+                let _: () = msg_send![ns_menu_item, setTag:child.id()];
+
+                // Store a raw pointer to the `MenuChild` as an instance variable on the native menu item
+                let ptr = Box::into_raw(Box::new(&*child));
+                (*ns_menu_item).set_ivar(BLOCK_PTR, ptr as usize);
+            }
+        }
+
```
amrbashir:
```suggestion
        if let PredfinedMenuItemType::About(_) = child.predefined_item_type {
            unsafe {
                let _: () = msg_send![ns_menu_item, setTarget: ns_menu_item];
                let _: () = msg_send![ns_menu_item, setTag:child.id()];

                // Store a raw pointer to the `MenuChild` as an instance variable on the native menu item
                let ptr = Box::into_raw(Box::new(&*child));
                (*ns_menu_item).set_ivar(BLOCK_PTR, ptr as usize);
            }
        }

```

src/predefined.rs
```diff
@@ -178,31 +178,65 @@ impl PredefinedMenuItem {
...
+    /// ## Platform-specific
+    ///
+    /// - **Windows / Linux:** This is ignored
+    pub credits: Option<String>,
+    /// The application icon.
+    ///
+    /// ## Platform-specific
+    ///
+    /// - **Windows / Linux:** This is ignored
+    pub icon: Option<Icon>,
```
amrbashir:
```suggestion
    ///
    /// ## Platform-specific
    ///
    /// - **macOS:** Unsupported.
    pub authors: Option<Vec<String>>,
    /// Application comments.
    ///
    /// ## Platform-specific
    ///
    /// - **macOS:** Unsupported.
    pub comments: Option<String>,
    /// The copyright of the application.
    pub copyright: Option<String>,
    /// The license of the application.
    ///
    /// ## Platform-specific
    ///
    /// - **macOS:** Unsupported.
    pub license: Option<String>,
    /// The application website.
    ///
    /// ## Platform-specific
    ///
    /// - **macOS:** Unsupported.
    pub website: Option<String>,
    /// The website label.
    ///
    /// ## Platform-specific
    ///
    /// - **macOS:** Unsupported.
    pub website_label: Option<String>,
    /// The credits.
    ///
    /// ## Platform-specific
    ///
    /// - **Windows / Linux:** Unsupported.
    pub credits: Option<String>,
    /// The application icon.
    ///
    /// ## Platform-specific
    ///
    /// - **Windows:** Unsupported.
    pub icon: Option<Icon>,
```

amrbashir (changes requested):

mxsdev:
EDIT: seems I was using `icon` instead of `logo` in `gtk::builders::AboutDialogBuilder`. `logo` is actually correct. Fixed in e3c1dfa 👍 

Having some issues with my implementation of icon for linux @amrbashir:

image

Am I initializing the image wrong? It works fine on osx.

```rust
            icon: Some(Icon::from_rgba(vec![255; 4 * 128 * 128], 128, 128).unwrap()),
```

Maybe gtk has a specific size requirement here?

src/predefined.rs
mxsdev: Added in 4cd70e5

src/platform_impl/macos/mod.rs
mxsdev: fd3891c

src/predefined.rs
mxsdev: 3e8c81a

amrbashir (approved):
Thanks for your contribution ❤️

PR #78, opened 2023-07-27
refactor!: use optional `Position` type and fallback to cursor pos

amrbashir:


lucasfernog (approved):
macOS works well even with a multi monitor setup 💯

amrbashir:
🎉

PR #84, opened 2023-08-03
fix(macos): manually retain/release NSMenu reference

lucasfernog:


src/platform_impl/macos/mod.rs
```diff
@@ -47,20 +47,42 @@ extern "C" {
...
+            let _: () = msg_send![self.ns_menu, retain];
+        }
+        Self {
+            id: self.id,
+            ns_menu: self.ns_menu.clone(),
+            children: self.children.clone(),
+        }
+    }
+}
+
```
amrbashir: we can remove this, we don't need it anymore, Clone is implemented for the outer type.
```suggestion
```

src/platform_impl/macos/mod.rs
```diff
@@ -311,6 +326,7 @@ impl MenuChild {
             id: COUNTER.next(),
             native_icon,
             accelerator,
+            children: None,
```
amrbashir: is there a reason for this? it should be covered by `..Default::default`, no?

lucasfernog: my bad, i was testing this compiler error i never faced https://doc.rust-lang.org/error_codes/E0509.html

PR #98, opened 2023-08-15
feat(windows): draw dark menubar if necessary

### LINKED ISSUE #97: Menu bar doesn't follow dark/light theme on Windows
The window title bar follows dark/light them on Windows nicely but not the menu bar.

Is it possible to do or a limitation of win32 api?

amrbashir:
closes #97

### DIFF
--- a/.changes/dark_menubar.md
+++ b/.changes/dark_menubar.md
@@ -1,0 +1,5 @@
+---
+"muda": "patch"
+---
+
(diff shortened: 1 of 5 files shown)


FabianLars:
What an insane change to "just" get a dark menu bar 😨 win32 api at its best 🙃

amrbashir:
it would have been much simpler if these APIs were documented but we gotta do with thi for now.

PR #109, opened 2023-08-30
refactor: remove unnecessary `Rc` usages

rhysd:
This PR removes `Rc` containers which are actually unnecessary.

Since `Rc` allocates its content and weak/strong counters on heap, it causes extra heap allocation. And accessing the content checks reference count which is an overhead.

In this crate, contents of `Rc` containers are actually shared by no one. They are always cloned. Such `Rc` containers can be removed safely. This does not introduce any breaking change because the `Rc` containers are implementation details.

### DIFF
--- a/src/items/check.rs
+++ b/src/items/check.rs
@@ -14,7 +14,7 @@ use crate::{accelerator::Accelerator, IsMenuItem, MenuId, MenuItemKind};
 /// [`Submenu`]: crate::Submenu
 #[derive(Clone)]
 pub struct CheckMenuItem {
-    pub(crate) id: Rc<MenuId>,
+    pub(crate) id: MenuId,
     pub(crate) inner: Rc<RefCell<crate::platform_impl::MenuChild>>,
 }
 
@@ -47,7 +47,7 @@ impl CheckMenuItem {
             None,
         );
         Self {
-            id: Rc::new(item.id().clone()),
+            id: item.id().clone(),
             inner: Rc::new(RefCell::new(item)),
         }
     }
@@ -65,7 +65,7 @@ impl CheckMenuItem {
     ) -> Self {
         let id = id.into();
         Self {
-            id: Rc::new(id.clone()),
+            id: id.clone(),
             inner: Rc::new(RefCell::new(crate::platform_impl::MenuChild::new_check(
                 text.as_ref(),
                 enabled,
--- a/src/items/icon.rs
+++ b/src/items/icon.rs
@@ -17,7 +17,7 @@ use crate::{
 /// [`Submenu`]: crate::Submenu
 #[derive(Clone)]
 pub struct IconMenuItem {
-    pub(crate) id: Rc<MenuId>,
+    pub(crate) id: MenuId,
     pub(crate) inner: Rc<RefCell<crate::platform_impl::MenuChild>>,
 }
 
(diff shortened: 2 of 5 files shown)


amrbashir (commented):
The `Rc` clone is cheaper than cloning the String and so it is not really unnecessary.

rhysd:
> The Rc clone is cheaper than cloning the String and so it is not really unnecessary.

Right. However if my understanding is correct, `Rc` for `id` field of each menu item is never cloned in this library. Their contents (`MenuId`) are cloned everywhere.

rhysd:
But OK, I understood your intention. I'll try to modify to keep `Rc` and share IDs.

rhysd:
For example, I can simply replace `Rc` with `Box` in `id` field of `MenuItem`. `Box` does not allow cloning itself so it means that the pointer is never cloned anywhere. Hence `MenuItem::id` field is never shared by anyone.

```diff
diff --git a/src/items/normal.rs b/src/items/normal.rs
index 52a79e8..edcbd3a 100644
--- a/src/items/normal.rs
+++ b/src/items/normal.rs
@@ -8,7 +8,7 @@ use crate::{accelerator::Accelerator, IsMenuItem, MenuId, MenuItemKind};
 /// [`Submenu`]: crate::Submenu
 #[derive(Clone)]
 pub struct MenuItem {
-    pub(crate) id: Rc<MenuId>,
+    pub(crate) id: Box<MenuId>,
     pub(crate) inner: Rc<RefCell<crate::platform_impl::MenuChild>>,
 }
 
@@ -30,7 +30,7 @@ impl MenuItem {
     pub fn new<S: AsRef<str>>(text: S, enabled: bool, acccelerator: Option<Accelerator>) -> Self {
         let item = crate::platform_impl::MenuChild::new(text.as_ref(), enabled, acccelerator, None);
         Self {
-            id: Rc::new(item.id().clone()),
+            id: Box::new(item.id().clone()),
             inner: Rc::new(RefCell::new(item)),
         }
     }
@@ -47,7 +47,7 @@ impl MenuItem {
     ) -> Self {
         let id = id.into();
         Self {
-            id: Rc::new(id.clone()),
+            id: Box::new(id.clone()),
             inner: Rc::new(RefCell::new(crate::platform_impl::MenuChild::new(
                 text.as_ref(),
                 enabled,
diff --git a/src/platform_impl/mod.rs b/src/platform_impl/mod.rs
index c6abb72..a681eee 100644
--- a/src/platform_impl/mod.rs
+++ b/src/platform_impl/mod.rs
@@ -47,7 +47,7 @@ impl MenuChild {
             MenuItemType::MenuItem => {
                 let id = c.borrow().id().clone();
                 MenuItemKind::MenuItem(MenuItem {
-                    id: Rc::new(id),
+                    id: Box::new(id),
                     inner: c,
                 })
             }
```

amrbashir:
Right, but iirc the usage of `Rc` was to make cloning `MenuItem`, `CheckMenuItem`...etc cheap so users can clone the items and move them into different closures if needed.

rhysd:
I see. So cloning the menu items is prioritized. I understood your intention.

amrbashir:
Correct, we had the users of tauri in-mind when designing this crate, and since there is a lot of closures in tauri, it was better to make the clone of menu items cheap as possible.

PR #112, opened 2023-08-31
feat: add `into_id` methods to all menu item types

### LINKED ISSUE #111: Allow converting a menu item into its menu ID
It is possible to access reference to menu item ID (e.g. `MenuItem::id()`) but it is not possible to convert menu item into its ID without clone.

In my application, I'm managing menu items by IDs and `std::collections::HashMap`. Here is a code example.

```rust
// Dispatch functions
fn open_file() {
    unimplemented!()
}
fn quit() {
    unimplemented!()
}

// Create menu
let open_file_item = MenuItem::new("Open File…", true, None);
let quit_item = MenuItem::new("Quit", true, None);
let file_menu = Submenu::with_items("&File", true, &[&open_file_item, &quit_item]);
let menu_bar = Menu::with_items(&[&file_menu])?;

// Create dispatch table
let menu_ids = HashMap::from_iter([
    (open_file_item.id().clone(), open_file),
    (quit_item.id().clone(),      quit),
]);

// ...

// Dispatch menu event
if let Ok(event) = self.receiver.try_recv() {
    if let Some(dispatch) = menu_ids.get(&event.id) {
        dispatch();
    }
}
```

Small issue in this code is that it needs to clone menu IDs. It is nice to have method to convert menu items into IDs.

```rust
let menu_ids = HashMap::from_iter([
    (open_file_item.into_id(), open_file),
    (quit_item.into_id(),      quit),
]);
```

rhysd:
Fixes #111

This PR implements `into_id` methods to `MenuItem`, `CheckMenuItem`, `PredefinedMenuItem`, and `Submenu`. It moves the menu item into its menu ID.

Since I'm not understanding the policy to add a new method to traits, I'm not sure `into_id` method should be added to `IsMenuItem` trait. If it should not, please let me know. I'll update this branch to move the trait methods to normal methods.

I didn't use `Rc::into_inner` because it was added at Rust 1.70 and Tauri's MSRV is earlier than the version.

### DIFF
--- a/src/items/check.rs
+++ b/src/items/check.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.inner
 // SPDX-License-Identifier: MIT
 
-use std::{cell::RefCell, rc::Rc};
+use std::{cell::RefCell, mem, rc::Rc};
 
 use crate::{accelerator::Accelerator, IsMenuItem, MenuId, MenuItemKind};
 
@@ -25,6 +25,15 @@ unsafe impl IsMenuItem for CheckMenuItem {
 
     fn id(&self) -> &MenuId {
         self.id()
+    }
+
+    fn into_id(mut self) -> MenuId {
+        // Note: `Rc::into_inner` is available from Rust 1.70
+        if let Some(id) = Rc::get_mut(&mut self.id) {
+            mem::take(id)
+        } else {
+            self.id().clone()
+        }
     }
 }
 
@@ -117,4 +126,13 @@ impl CheckMenuItem {
     pub fn set_checked(&self, checked: bool) {
         self.inner.borrow_mut().set_checked(checked)
     }
+}
+
+#[test]
+fn test_from_id_and_into_id() {
+    let id = "TEST ID".to_string();
+    let item = CheckMenuItem::with_id(&id, "test", true, true, None);
+    let expected = MenuId(id);
+    assert_eq!(item.id(), &expected);
+    assert_eq!(item.into_id(), expected);
 }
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -256,6 +256,8 @@ pub unsafe trait IsMenuItem {
     fn kind(&self) -> MenuItemKind;
 
     fn id(&self) -> &MenuId;
+
+    fn into_id(self) -> MenuId;
```
amrbashir: It is fine to add this to `IsMenuItem` trait but the trait is meant for internal usages only, so you also need to add it to each struct and also add it to `MenuItemKind` enum

src/items/check.rs
```diff
@@ -118,3 +127,12 @@ impl CheckMenuItem {
...
 }
+
+#[test]
+fn test_from_id_and_into_id() {
+    let id = "TEST ID".to_string();
+    let item = CheckMenuItem::with_id(&id, "test", true, true, None);
+    let expected = MenuId(id);
+    assert_eq!(item.id(), &expected);
+    assert_eq!(item.into_id(), expected);
+}
```
amrbashir: move these tests to `src/items/mod.rs`

amrbashir (changes requested):

rhysd:
Thank you for the review. I will handle them tomorrow.

src/lib.rs
amrbashir: Actually, since there isn't any usage for it internally, it is useless to add it here

rhysd:
@amrbashir I fixed your review comments at bcd33b1.

- Smaller trait is better. Since `into_id` was unnecessary for any implementation, I removed the method from the `IsMenuItem` trait
- Moved all `into_id` trait methods to each structs
- Added `MenuItemKind::into_id` method
- Gathered all tests into a single test case at `src/items/mod.rs`

amrbashir (approved):

PR #120, opened 2023-09-12
fix(macOS): fixed over release the separator

pewsheen:
Issue: https://github.com/tauri-apps/tauri/issues/7760

I think the separator is already an autorelease object since we didn't `alloc` and `init` by ourselves. So, calling the `autorelease()` again will cause an over-release and we should not release it manually.

### DIFF
--- a/.changes/fix-separator-autorelease
+++ b/.changes/fix-separator-autorelease
@@ -1,0 +1,5 @@
+---
+"muda": "patch"
+---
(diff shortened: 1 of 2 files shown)


amrbashir:
@lucasfernog can you review this one?

lucasfernog (approved):
fixed the issue, nice one!

PR #131, opened 2023-10-17
fix: fix clippy warnings

rhysd:
Recently clippy enabled `clippy::unwrap_or_default` rule by default and warnings reported by the rule made CI fail. This PR fixes them.


amrbashir (approved):
Thanks

PR #133, opened 2023-10-19
chore(deps): upgrade gtk to 0.18

dklassic:
- [x] Waiting for `wry` 0.34

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,5 +51,5 @@ png = "0.17"
 [dev-dependencies]
 winit = "0.28"
 tao = "0.23"
-wry = "0.34
+wry = "0.34"
 image = "0.24"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -40,9 +40,9 @@ features = [
 ]
 
 [target."cfg(target_os = \"linux\")".dependencies]
-gtk = "0.16"
-gdk = "0.16"
-gdk-pixbuf = "0.16"
+gtk = "0.18"
+gdk = "0.18"
+gdk-pixbuf = "0.18"
```
amrbashir: We should probably remove these and just use the re-exported modules in `gtk` crate. I plan to do the same for all the crates we maintain btw, so it is better to start now.

amrbashir (changes requested):
Ah you beat me to it, thanks.

Cargo.toml
dklassic: Makes sense, I'll do some refactoring.

amrbashir (approved):

PR #138, opened 2023-11-10
ci: add running test with miri to CI

pcwizz:
For T1 Rust targets Miri is able to check for several forms of undefined behavior that rustc can not.

### DIFF
--- a/src/items/mod.rs
+++ b/src/items/mod.rs
@@ -19,6 +19,7 @@ mod test {
     use crate::{CheckMenuItem, IconMenuItem, MenuId, MenuItem, PredefinedMenuItem, Submenu};
 
     #[test]
+    #[cfg_attr(all(miri, not(target_os = "linux")), ignore)]
     fn it_returns_same_id() {
         let id = MenuId::new("1");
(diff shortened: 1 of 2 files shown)


src/items/mod.rs
```diff
@@ -19,6 +19,7 @@ mod test {
     use crate::{CheckMenuItem, IconMenuItem, MenuId, MenuItem, PredefinedMenuItem, Submenu};
 
     #[test]
+    #[cfg_attr(all(miri, not(target_os = "linux")), ignore)]
```
amrbashir, line 22: why does these examples need to be ignored on Windows and macOS using miri?

pcwizz: Because on MacOS and windows these tests involve FFI which is not supported by Miri at this time.

MacOS: https://github.com/pcwizz/muda/actions/runs/6826919902/job/18567898544#step:7:181
Windows: https://github.com/pcwizz/muda/actions/runs/6826919902/job/18567898285#step:7:124

amrbashir: I see, thanks

amrbashir (approved):

PR #159, opened 2024-03-09
hack addressing tauri issue #9127 (context menus overwrite WNDPROC dwrefdata on windows)

gulbanana:
See https://github.com/tauri-apps/muda/issues/162 for more details.
This is not a good fix, but it demonstrates the nature of the problem.


amrbashir:
Thank you, I have opened https://github.com/tauri-apps/muda/pull/166 to fix this behavior.

PR #214, opened 2024-08-17
fix(windows): use `TPM_RETURNCMD` instead of subclass for context menus

### LINKED ISSUE #204: show_context_menu_for_hwnd() requires self to last for the lifetime of the pop up
`Menu::show_context_menu_for_hwnd()` takes `&self` and returns immediately.  If `&self` is dropped or moved, the act of selecting a menu item will result in a stray memory access.

Shouldn't this be changed to something safe (e.g. - taking `Rc<Menu>`) or be marked as `unsafe`?  Though I suspect that anybody using `muda` and messing around with `HWND`s is in any case "'going offroad'"

npwoods:
These changes are to address a design flaw in how `muda` handles context menus on Windows.

The normal way menus with `muda` are used on Windows is they get attached to a window, and the window is subclassed to intercept commands (`WM_COMMAND`) and some drawing behaviors (I'm not 100% sure on this, but this appears in place to support theming and dark mode).  When the `Menu` is dropped, it will be unattached from all windows and the subclasses will be removed.

Context menus seem to follow a slightly different pattern.  When `show_context_menu_for_hwnd()` is called, the subclass is attached to the window, but never removed.  This behavior is problematic because the menu can be dropped without the subclass removed, causing a crash.  Furthermore, having this subclass attached unnecessarily is not necessarily a good idea.

These changes to `muda` refactor menu support on Windows to ensure that the subclass (which seems to be necessary to support theming) lives only while the context menu is active, which should address these problems.

These changes also address a memory leak where when an HWND is subclassed, the reference to the menu object was placed in a `Box` seemingly unnecessarily, and this `Box` never seems to be dropped.  This was changed to using straight references.

### DIFF
--- a/.changes/win32-context-menu-fixes
+++ b/.changes/win32-context-menu-fixes
@@ -2,12 +2,4 @@
 "muda": "patch"
 ---
 
-These changes are to address a design flaw in how `muda` handles context menus on Windows.
-    
-The normal way menus with `muda` are used on Windows is they get attached to a window, and the window is subclassed to intercept commands (`WM_COMMAND`) and some drawing behaviors (I'm not 100% sure on this, but this appears in place to support theming and dark mode).  When the `Menu` is dropped, it will be unattached from all windows and the subclasses will be removed.
-    
-Context menus seem to follow a slightly different pattern.  When `show_context_menu_for_hwnd()` is called, the subclass is attached to the window, but never removed.  This behavior is problematic because the menu can be dropped without the subclass removed, causing a crash.  Furthermore, having this subclass attached unnecessarily is not necessarily a good idea.
-    
-These changes to `muda` refactor menu support on Windows to remove the HWND subclassing when the context menu is active.
-    
-These changes also address a memory leak where when an HWND is subclassed, the reference to the menu object was placed in a `Box` seemingly unnecessarily, and this `Box` never seems to be dropped.  This was changed to using straight references.
+On Windows, fix crash when showing a context menu but dropping the Menu before the context menu is closed.
--- a/.changes/win32-context-menu-fixes
+++ b/.changes/win32-context-menu-fixes
@@ -1,0 +1,13 @@
+---
+"muda": "patch"
+---
+
+These changes are to address a design flaw in how `muda` handles context menus on Windows.
+    
+The normal way menus with `muda` are used on Windows is they get attached to a window, and the window is subclassed to intercept commands (`WM_COMMAND`) and some drawing behaviors (I'm not 100% sure on this, but this appears in place to support theming and dark mode).  When the `Menu` is dropped, it will be unattached from all windows and the subclasses will be removed.
+    
+Context menus seem to follow a slightly different pattern.  When `show_context_menu_for_hwnd()` is called, the subclass is attached to the window, but never removed.  This behavior is problematic because the menu can be dropped without the subclass removed, causing a crash.  Furthermore, having this subclass attached unnecessarily is not necessarily a good idea.
+    
+These changes to `muda` refactor menu support on Windows to remove the HWND subclassing when the context menu is active.
+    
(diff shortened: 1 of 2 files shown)


src/platform_impl/windows/mod.rs
```diff
@@ -413,16 +413,27 @@ impl Menu {
...
         }
-        show_context_menu(hwnd as _, hpopupmenu, position)
+        let rc = show_context_menu(hwnd as _, self.hpopupmenu, position);
+        unsafe {
+            RemoveWindowSubclass(
+                hwnd as _,
+                Some(menu_subclass_proc),
+                CONTEXT_MENU_SUBCLASS_ID,
+            );
+        }
```
amrbashir: Not needed, we can remove this and `SetWindowSubclass` before calling `show_context_menu`.

src/platform_impl/windows/mod.rs
```diff
@@ -924,16 +935,27 @@ impl MenuChild {
...
-                Box::into_raw(Box::new(self)) as _,
+                dwrefdata_from_obj(self),
+            );
+        }
+        let rc = show_context_menu(hwnd as _, self.hpopupmenu, position);
+        unsafe {
+            RemoveWindowSubclass(
+                hwnd as _,
+                Some(menu_subclass_proc),
+                CONTEXT_SUBMENU_SUBCLASS_ID,
```
amrbashir: same here

amrbashir (changes requested):
Thank you, just a few comments and this should be good to go. 

Could you also add a change file in `.changes` directory?

npwoods:
I am interpreting your feedback as "'Now that the subclass doesn't handle `WM_COMMAND` for context menus, we can stop subclassing context menus'".  I'm surprised to hear this, as I thought that the subclass was still necessary for dark mode.  100% your call of course!

In any case, I've submitted my changes.  If there is anything else I need to do, let me know!

amrbashir:
> I am interpreting your feedback as "'Now that the subclass doesn't handle `WM_COMMAND` for context menus, we can stop subclassing context menus'".

Yeah that's what I meant exactly

> I'm surprised to hear this, as I thought that the subclass was still necessary for dark mode. 100% your call of course!

The subclass only draws the menu bar (the strip on top of the hwnd), context menus and also submenus from the menu bar are not drawn by us in dark mode.

.changes/win32-context-menu-fixes
```diff
@@ -0,0 +1,13 @@
...
+    
+The normal way menus with `muda` are used on Windows is they get attached to a window, and the window is subclassed to intercept commands (`WM_COMMAND`) and some drawing behaviors (I'm not 100% sure on this, but this appears in place to support theming and dark mode).  When the `Menu` is dropped, it will be unattached from all windows and the subclasses will be removed.
+    
+Context menus seem to follow a slightly different pattern.  When `show_context_menu_for_hwnd()` is called, the subclass is attached to the window, but never removed.  This behavior is problematic because the menu can be dropped without the subclass removed, causing a crash.  Furthermore, having this subclass attached unnecessarily is not necessarily a good idea.
+    
+These changes to `muda` refactor menu support on Windows to remove the HWND subclassing when the context menu is active.
+    
+These changes also address a memory leak where when an HWND is subclassed, the reference to the menu object was placed in a `Box` seemingly unnecessarily, and this `Box` never seems to be dropped.  This was changed to using straight references.
```
amrbashir: This is too much information for a changelog entry, it should be something simple like 

```suggestion
On Windows, fix crash when showing a context menu but dropping the Menu before the context menu is closed.
```

amrbashir (approved):

amrbashir:
Thank you

npwoods:
Thank you for all your help, and helping me muddle through the details!

Issue #231, opened 2024-10-01
Keys are not being recognized


0PandaDEV:
```bash
Error registering shortcut: GlobalHotkey("Couldn't recognize \"ControlLeft\" as a valid key for hotkey, if you feel like it should be, please report this to https://github.com/tauri-apps/muda")
```

Im on a M1 Pro. In the [docs.rs](https://docs.rs/tauri-plugin-global-shortcut/2.0.0-rc.2/tauri_plugin_global_shortcut/enum.Code.html) it says that this key exists.

I have the same problem with all the modifier keys ControlLeft, ControlRight, MetaLeft, MetaRight, ShiftLeft, ShiftRight, AltLeft, AltRight.

0PandaDEV:
?

amrbashir:
These keys are modifiers and can't be used as keys.

PR #232, opened 2024-10-01
Fixed Modifier Left, Right keys

0PandaDEV:
I don't know how I can test these changes because I use it through the global-shortcut plugin from tauri normally.


amrbashir:
as explained in https://github.com/tauri-apps/muda/issues/231 these are modifiers and can't be used as keys.

PR #236, opened 2024-10-25
feat: return context menu dismiss result

pewsheen:
- Adding context menu dismiss result, so we would know the menu is dismissed because the user selected an item or canceled it.

- Windows, macOS
  -  Return `true` when item selected
  - Return `false` when the menu is canceled (including clicking on empty space, pressing ESC)
- Linux (Wayland)
  - Return `true` when an item is selected or clicking on empty space
  - Return `false` when the menu is canceled (like hitting ESC)

### DIFF
--- a/.changes/add-context-menu-dismiss-result.md
+++ b/.changes/add-context-menu-dismiss-result.md
@@ -2,4 +2,4 @@
 "muda": patch
 ---
 
-On macOS and Windows, context menu returns it's dismissed result.
+Return `bool` in `ContextMenu::show_context_menu_for_hwnd` on Windows and `ContextMenu::show_context_menu_for_nsview` on macOS to indicate why the context menu was closed.
--- a/.changes/add-context-menu-dismiss-result.md
+++ b/.changes/add-context-menu-dismiss-result.md
@@ -1,5 +1,5 @@
 ---
-"muda": patch
+"muda": minor
 ---
 
-Return `bool` in `ContextMenu::show_context_menu_for_hwnd` on Windows and `ContextMenu::show_context_menu_for_nsview` on macOS to indicate why the context menu was closed.
+Return `bool` in `ContextMenu::show_context_menu_for_hwnd`, `ContextMenu::show_context_menu_for_nsview` and `ContextMenu::show_context_menu_for_gtk_window` to indicate why the context menu was closed.
--- a/.changes/add-context-menu-dismiss-result.md
+++ b/.changes/add-context-menu-dismiss-result.md
@@ -1,0 +1,5 @@
+---
+"muda": patch
+---
+
+On macOS and Windows, context menu returns it's dismissed result.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -311,11 +311,17 @@ pub trait ContextMenu {
     ///
     /// - `position` is relative to the window top-left corner, if `None`, the cursor position is used.
     ///
+    /// Returns `true` if menu tracking ended because an item was selected, and `false` if menu tracking was cancelled for any reason.
+    ///
     /// # Safety
     ///
     /// The `hwnd` must be a valid window HWND.
     #[cfg(target_os = "windows")]
-    unsafe fn show_context_menu_for_hwnd(&self, hwnd: isize, position: Option<dpi::Position>);
+    unsafe fn show_context_menu_for_hwnd(
+        &self,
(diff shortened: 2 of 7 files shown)


.changes/add-context-menu-dismiss-result.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": patch
+---
+
+On macOS and Windows, context menu returns it's dismissed result.
```
amrbashir:
```suggestion
Return `bool` in `ContextMenu::show_context_menu_for_hwnd` on Windows and `ContextMenu::show_context_menu_for_nsview` on macOS to indicate why the context menu was closed.
```

amrbashir (changes requested):
Can't we implement the same on Linux using the signals emitted on menus? 
see https://docs.gtk.org/gtk3/signal.MenuShell.selection-done.html and https://docs.gtk.org/gtk3/signal.MenuShell.cancel.html

pewsheen:
> Can't we implement the same on Linux using the signals emitted on menus? see https://docs.gtk.org/gtk3/signal.MenuShell.selection-done.html and https://docs.gtk.org/gtk3/signal.MenuShell.cancel.html

Hmm, I'm expecting that the result could return synchronously when it is dismissed. Signals seem unable to do it synchronously.
If we need to support Linux, maybe it should emit a dismiss `MenuEvent` on all platforms?

amrbashir:
we can probably use a channel to synchronously wait for a result, no?

pewsheen:
> we can probably use a channel to synchronously wait for a result, no?

Cool, I'll try it later

amrbashir:
thanks, you may also need to do something like this https://github.com/tauri-apps/wry/blob/1d63fa325327a02a0a8be9ee50ce1eb7a0e8e04f/src/webkitgtk/mod.rs#L874-L880 in case it is dead-locking

pewsheen:
> Can't we implement the same on Linux using the signals emitted on menus? see https://docs.gtk.org/gtk3/signal.MenuShell.selection-done.html and https://docs.gtk.org/gtk3/signal.MenuShell.cancel.html

Just added these two signals.
But the events are slightly not as expected that when I mouse click on an empty space. On macOS and Windows, they will see this behavior as canceled, but on Linux (at least on Wayland), it sees it as select-done

src/lib.rs
```diff
@@ -341,8 +347,16 @@ pub trait ContextMenu {
     /// Shows this menu as a context menu inside a [`gtk::Window`]
     ///
     /// - `position` is relative to the window top-left corner, if `None`, the cursor position is used.
+    ///
+    /// Returns `true` if menu tracking ended because an item was selected or clicked other place to dismiss the menu
```
amrbashir:
```suggestion
    /// Returns `true` if menu tracking ended because an item was selected or clicked outside the menu to dismiss it.
```

.changes/add-context-menu-dismiss-result.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": patch
+---
+
+Return `bool` in `ContextMenu::show_context_menu_for_hwnd` on Windows and `ContextMenu::show_context_menu_for_nsview` on macOS to indicate why the context menu was closed.
```
amrbashir:
```suggestion
---
"muda": minor
---

Return `bool` in `ContextMenu::show_context_menu_for_hwnd`, `ContextMenu::show_context_menu_for_nsview` and `ContextMenu::show_context_menu_for_gtk_window` to indicate why the context menu was closed.
```

amrbashir (approved):

amrbashir (approved):

src/platform_impl/gtk/mod.rs
```diff
@@ -1411,14 +1411,40 @@ fn show_context_menu(
             }
         }
 
+        let (tx, rx) = crossbeam_channel::unbounded();
+        let tx_clone = tx.clone();
+        let id = gtk_menu.connect_cancel(move |_| tx_clone.send(false).unwrap_or(()));
+        let id2 = gtk_menu.connect_selection_done(move |s| tx.send(true).unwrap_or(()));
```
amrbashir:
```suggestion
        let id2 = gtk_menu.connect_selection_done(move |_| tx.send(true).unwrap_or(()));
```

amrbashir (approved):

PR #239, opened 2024-11-01
feat: add snapshot APIs

dfaust:
I had to expose a few internals and add some image conversion functions to make muda work with my ksni patch for the tray-icon crate.

Companion PR for tray-icon:
Replace libappindicator with ksni: https://github.com/tauri-apps/tray-icon/pull/201

Tauri issue:
[feat] Use ksni crate for tray icons on Linux: https://github.com/tauri-apps/tauri/issues/11293


src/icon.rs
```diff
@@ -163,6 +163,18 @@ impl Icon {
...
+    #[cfg(target_os = "linux")]
+    pub fn to_pixbuf(&self) -> gtk::gdk_pixbuf::Pixbuf {
+        self.inner.to_pixbuf()
+    }
+
+    /// Convert the icon into PNG.
+    #[cfg(target_os = "linux")]
+    pub fn to_png(&self) -> Vec<u8> {
+        self.inner.to_png()
+    }
```
amrbashir: maybe we can move this method into `tray-icon` crate, and only expose `to_pixbuf`

src/items/icon.rs
```diff
@@ -145,24 +145,24 @@ impl IconMenuItem {
         &self.id
     }
 
-    /// Get the text for this check menu item.
+    /// Get the text for this menu item.
```
amrbashir:
```suggestion
    /// Get the text for this icon menu item.
```

src/items/icon.rs
```diff
@@ -145,24 +145,24 @@ impl IconMenuItem {
...
     pub fn text(&self) -> String {
         self.inner.borrow().text()
     }
 
-    /// Set the text for this check menu item. `text` could optionally contain
+    /// Set the text for this menu item. `text` could optionally contain
```
amrbashir:
```suggestion
    /// Set the text for this icon menu item. `text` could optionally contain
```

src/items/icon.rs
```diff
@@ -145,24 +145,24 @@ impl IconMenuItem {
...
     /// an `&` before a character to assign this character as the mnemonic
-    /// for this check menu item. To display a `&` without assigning a mnemenonic, use `&&`.
+    /// for this menu item. To display a `&` without assigning a mnemenonic, use `&&`.
```
amrbashir:
```suggestion
    /// for this icon menu item. To display a `&` without assigning a mnemenonic, use `&&`.
```

src/items/icon.rs
```diff
@@ -145,24 +145,24 @@ impl IconMenuItem {
...
     pub fn set_text<S: AsRef<str>>(&self, text: S) {
         self.inner.borrow_mut().set_text(text.as_ref())
     }
 
-    /// Get whether this check menu item is enabled or not.
+    /// Get whether this menu item is enabled or not.
```
amrbashir:
```suggestion
    /// Get whether this icon menu item is enabled or not.
```

src/items/icon.rs
```diff
@@ -145,24 +145,24 @@ impl IconMenuItem {
...
     pub fn is_enabled(&self) -> bool {
         self.inner.borrow().is_enabled()
     }
 
-    /// Enable or disable this check menu item.
+    /// Enable or disable this menu item.
```
amrbashir:
```suggestion
    /// Enable or disable this icon menu item.
```

src/items/icon.rs
```diff
@@ -172,6 +172,11 @@ impl IconMenuItem {
         self.inner.borrow_mut().set_accelerator(accelerator)
     }
 
+    /// Get the icon for this menu item.
```
amrbashir:
```suggestion
    /// Get the icon for this icon menu item.
```

src/items/predefined.rs
```diff
@@ -241,7 +246,7 @@ fn test_about_metadata() {
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 #[allow(clippy::large_enum_variant)]
-pub(crate) enum PredefinedMenuItemType {
+pub enum PredefinedMenuItemType {
```
amrbashir: if we are to expose this type, I think it should be called `PredefinedMenuItemKind`

src/lib.rs
```diff
@@ -350,6 +353,9 @@ pub trait ContextMenu {
     #[cfg(target_os = "linux")]
     fn gtk_context_menu(&self) -> gtk::Menu;
 
+    /// Get all menu items as a vector of `MenuItemKind` elements.
```
amrbashir:
```suggestion
    /// Get all menu items within this context menu.
```

amrbashir (changes requested):

amrbashir (approved):

src/items/compat.rs
```diff
@@ -0,0 +1,119 @@
...
+                    .icon()
+                    .map(|icon| icon.to_pixbuf().save_to_bufferv("png", &[]).unwrap()),
+                predefined_menu_item_kind: None,
+            }
+            .into(),
+        }
+    }
+}
+
+pub fn strip_accelerator(text: impl AsRef<str>) -> String {
```
amrbashir: is this meant to remove mnemonic? if so, we need to keep `&&`

src/items/icon.rs
```diff
@@ -145,48 +194,89 @@ impl IconMenuItem {
...
+    #[cfg(target_os = "macos")]
+    pub fn set_native_icon(&self, icon: Option<NativeIcon>) {
+        let mut item = self.inner.borrow_mut();
+        item.set_native_icon(icon);
+
+        #[cfg(feature = "ksni")]
+        self.compat.store(Arc::new(Self::compat_menu_item(&item)));
+        
+        #[cfg(feature = "ksni")]
+        crate::send_menu_update();
```
amrbashir: why run these on macOS? shouldn't it be on Linux only? I believe reverting this block should be enough

src/lib.rs
```diff
@@ -417,11 +445,24 @@ impl MenuEvent {
...
+    pub fn send(event: MenuEvent) {
         if let Some(handler) = MENU_EVENT_HANDLER.get_or_init(|| None) {
             handler(event);
         } else {
             let _ = MENU_CHANNEL.0.send(event);
         }
     }
 }
+
+#[cfg(feature = "ksni")]
```
amrbashir: these should also be cfg'd behind linux-only target and same for all related logic

amrbashir (commented):
Apologies for the long delay

src/items/compat.rs
```diff
@@ -0,0 +1,119 @@
...
+                enabled: icon_menu_item.is_enabled(),
+                icon: icon_menu_item
+                    .icon()
+                    .map(|icon| icon.to_pixbuf().save_to_bufferv("png", &[]).unwrap()),
+                predefined_menu_item_kind: None,
+            }
+            .into(),
+        }
+    }
+}
```
amrbashir: This implementation seems to not be used at all, let's remove

src/items/compat.rs
dfaust: Do we have to keep `&&`, or turn it into `&`?

Btw. the `.replace("[~~]", "&")` in `to_gtk_mnemonic` is redundant.

src/items/icon.rs
dfaust: Probably a copy-paste mistake...

src/lib.rs
dfaust: I can do that. But it feels a bit redundant.

amrbashir: it should result in faster compile times on other platforms, no?

src/items/compat.rs
amrbashir: > Do we have to keep `&&`, or turn it into `&`?

it should turn into `&`

> Btw. the `.replace("[~~]", "&")` in `to_gtk_mnemonic` is redundant.

actually that one is correct, `.replace("[~~]", "&&")` is the redundant one

src/items/icon.rs
```diff
@@ -145,48 +194,83 @@ impl IconMenuItem {
...
     /// ## Platform-specific:
     ///
     /// - **Windows / Linux**: Unsupported.
-    pub fn set_native_icon(&self, _icon: Option<NativeIcon>) {
-        #[cfg(target_os = "macos")]
-        self.inner.borrow_mut().set_native_icon(_icon)
+    #[cfg(target_os = "macos")]
+    pub fn set_native_icon(&self, icon: Option<NativeIcon>) {
+        let mut item = self.inner.borrow_mut();
+        item.set_native_icon(icon);
```
amrbashir: let's use a single function so documentation is not duplicated 
```suggestion
    pub fn set_native_icon(&self, _icon: Option<NativeIcon>) {
        #[cfg(target_os = "macos")]
        self.inner.borrow_mut().set_native_icon(_icon)
```

src/items/icon.rs
```diff
@@ -145,48 +194,83 @@ impl IconMenuItem {
...
     }
 
+    /// Change this menu item icon to a native image or remove it.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// - **Windows / Linux**: Unsupported.
+    #[cfg(not(target_os = "macos"))]
+    pub fn set_native_icon(&self, _icon: Option<NativeIcon>) {}
+
```
amrbashir: and then let's remove this

src/lib.rs
dfaust: You're right. I didn't notice that the ksni feature is enabled on all platforms.
I added the target_os cfg everywhere, except `platform_impl/gtk`.

amrbashir (approved):
Thanks @dfaust. Apologies for the very long wait, and apologies for taking over your PRs as I wasn't too happy with the direction and I had to refactor the crate a bit in https://github.com/tauri-apps/muda/pull/394

dfaust:
@amrbashir Thanks for taking over. I don't have the time for this right now.

PR #244, opened 2024-11-05
feat: add `Accelerator::key` and `Accelerator::modifiers` getters

stijnfrishert:
This is a PR to expose the `Modifiers` and `Code` fields within an `Accelerator`, immutably. The types are public anyway, so I don't see why you shouldn't be able to inspect what you set earlier.

I personally need this to extract the modifiers and code and fill a human readable struct with them like so. I'm only using the `Accelerator` type from `muda` because it nicely represents hotkeys and supports string parsing.

```rust
struct Hotkey {
  shift: bool,
  alt: bool,
  control: bool,
  super: bool,
  key: Code,
}
```

### DIFF
--- a/.changes/expose-modifiers-and-key-on-accelerator.md
+++ b/.changes/expose-modifiers-and-key-on-accelerator.md
@@ -1,0 +1,5 @@
+---
+"muda": minor
+---
+
+Expose `modifiers()` and `key()` on `Accelerator` immutably, to be able to inspect what was set on an `Accelerator`.
--- a/.changes/expose-modifiers-and-key-on-accelerator.md
+++ b/.changes/expose-modifiers-and-key-on-accelerator.md
@@ -2,4 +2,4 @@
 "muda": minor
 ---
 
-Expose `modifiers()` and `key()` on `Accelerator` immutably, to be able to inspect what was set on an `Accelerator`.
+Add `Accelerator::modifiers` and `Accelerator::key` getter methods.
--- a/.changes/expose-modifiers-and-key-on-accelerator.md
+++ b/.changes/expose-modifiers-and-key-on-accelerator.md
@@ -1,5 +1,5 @@
 ---
-"muda": minor
+"muda": patch
 ---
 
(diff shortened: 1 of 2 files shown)


src/accelerator.rs
```diff
@@ -98,6 +98,16 @@ impl Accelerator {
...
 
+    /// Returns the modifier for this accelerator
+    pub fn modifiers(&self) -> &Modifiers {
+        &self.mods
+    }
+
+    /// Returns the code for this accelerator
+    pub fn key(&self) -> &Code {
+        &self.key
+    }
```
amrbashir: `Modifiers` and `Code` are copy types so we can just return them without a reference.

```suggestion
    /// Returns the modifier for this accelerator
    pub fn modifiers(&self) -> Modifiers {
        self.mods
    }

    /// Returns the code for this accelerator
    pub fn key(&self) -> Code {
        self.key
    }
```

amrbashir (changes requested):
Could you also add a change file in `.changes` directory?

stijnfrishert:
Amended my commit and pushed the new changes, including a file in `.changes`, @amrbashir 👍

.changes/expose-modifiers-and-key-on-accelerator.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": minor
+---
+
+Expose `modifiers()` and `key()` on `Accelerator` immutably, to be able to inspect what was set on an `Accelerator`.
```
amrbashir:
```suggestion
Add `Accelerator::modifiers` and `Accelerator::key` getter methods.
```

.changes/expose-modifiers-and-key-on-accelerator.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": minor
```
amrbashir:
```suggestion
"muda": patch
```

amrbashir (approved):
Thank you

PR #247, opened 2024-11-22
fixed + and - accelerators on macOS

platevoltage:
NumpadAdd and NumpadSubtract accelerators were added to the MacOS accelerator definitions. Command+ and Command- are common shortcuts for zooming on MacOS.


amrbashir:
Thank you, this will be fixed in #385

PR #272, opened 2025-02-01
feat!: gtk4

### LINKED ISSUE #270: Upgrade `muda` to `gtk4-rs`
We need to upgrade from [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) to migrate a unsoundness [issue](https://github.com/advisories/GHSA-wrw7-89jp-8q8g) found in glib < 0.2.
Please claim this issue if you want to support or work on this and keep in mind this is a prioritized issue and we try to migrate as soon as possible. If the issue is already assigned to someone you can either reach out with comments or questions reach out for collaboration to the assigned person.

This transition includes:

- [ ]  Upgrading [`gtk`](https://crates.io/crates/gtk) to [`gtk4-rs`](https://crates.io/crates/gtk4) and handling all errors caused by the dependency change
- [ ]  Handle all changes in a backwards compatible way if possible
    - [ ]  Raise awareness to the working group if breaking changes are required
- [ ]  Document migration errors that may be issues in the future or which were complicated to handle
- [ ]  Reach out to the working group (alternatively @tweidinger first) if blockers or upstream issues are hit

amrbashir:
closes https://github.com/tauri-apps/muda/issues/270

kicking off work on gtk4 migration

help is appreciated


rmakestrash-jpg:
Hey @amrbashir! Thanks for kicking off the GTK4 migration work - I used it as the foundation for my project's GTK4 support and it's been working great.

**My muda fork:** https://github.com/rmakestrash-jpg/muda/tree/feat/gtk4

**Changes on top of your work:**
- Fixed bug in `accelerator.rs`: `Code::Digit0` was incorrectly mapped to `"1"` instead of `"0"`
- Added auto-sync workflow to stay up-to-date with upstream

Tested on Wayland and everything works great!

I used Claude to help with this. Would really appreciate if you could review the work and let me know if there's anything I can help with to get this merged upstream!

johncarmack1984:
Pulled the branch, ran a plain gtk4 example app, mostly trying to break the menus. Rev 74c67ff, GTK 4.18.6 on Debian trixie, headless X11 (Xvfb, no WM), menubar and context menus driven by keyboard and mouse via xdotool, mutating items on a timer while menus were open.

Worked perfectly:
- icon items render
- set_text, set_icon, set_checked, and set_enabled all update live in an open popover
- set_accelerator after init updates the label
- the same item in two menus, or twice in one menu, renders and stays in sync
- insert and remove after init work
- show_context_menu_for_gtk_window returns true/false correctly, survives a dozen consecutive shows with the window's descendant widget count flat, and works on a new window after the original parent was destroyed
- Quit and About
- F10/arrow/Return navigation, mnemonics, and closed-menu accelerators all fire with the right event ids. A 250-item submenu attaches in ~19ms.

What broke:

- Accelerators with punctuation or numpad keys are silently dropped. `key_to_gtk` passes printable characters through verbatim, but `gtk_accelerator_parse` wants keysym names for punctuation: `Code::Minus` becomes `<Control>-` and fails with ``Gtk-CRITICAL: Unable to parse accelerator '<Control>-'``, so nothing installs and no label renders.
- Same for NumpadAdd (`<Control>+`, and the KP_ distinction is lost). Letters, digits, and space are fine.
- The `_ => String::new()` arm fails the same way for unmapped keys, as a bare `<Control>`. Mapping punctuation to keysym names (minus, plus, equal, etc.) fixes it. Happy to send a PR for that or leave it with you.
- A separator inside a nested submenu reserves its space but never draws: menubar > submenu built as [item, separator, item] shows the gap with no line. Top-level separators are fine, so it looks like the custom-widget binding misses lazily-created submenu popovers.
- Submenu::set_icon doesn't render anywhere I could find. IconMenuItem icons are fine.

Change adjacent / "Relevant But Not Exactly the Menus":
- predefined Copy/Cut/Paste vanish silently rather than rendering disabled, which stock Edit menus will feel.
- On macOS the winit_gtk4 example isn't cfg-gated, so `clippy --all-targets` fails before anything else gets a chance
- libgtk-4-dev missing on the runners, looks like most of the CI red.

Happy to re-run any of this as the branch moves.

PR #283, opened 2025-03-05
Make gtk an optional feature

### LINKED ISSUE #282: Make GTK an optional dependency
This might be a bit controversial, but I'd like to propose making GTK an optional dependency (enabled by default).

The motivation behind this request is our use case at Slint. We use the Muda crate on Windows and macOS, but not on Linux. However, due to how Cargo handles dependencies, all platform-specific dependencies are included in Cargo.lock, even if they can't possibly be used on any platform since the platforms don't overlap.

Currently, muda still depends on the old gtk3 crate, which has a security advisory. As a result, all our users receive security advisory notifications, even though we don't use Muda on Linux. This happens because the insecure dependency is still pulled into Cargo.lock. See https://github.com/slint-ui/slint/issues/7800 for more details.

I understand that you're planning to upgrade to GTK4, but that will take time. Even after the upgrade, it would be beneficial to have the option to disable GTK entirely, as it comes with many dependencies, and similar issues could arise in the future.

I also understand that most Linux users of this crate will want GTK, so if you’re not open to this change, feel free to close the issue. However, if you’re open to making GTK optional, I’d be happy to contribute a PR.

Looking forward to your thoughts. Please let me know if you have another idea on how to solve the problem.

ogoffart:
This helps projects that uses muda on platform other than Linux to not get the gtk dependencies in their Cargo.lock (and the security advisories that goes with it)

Fixes https://github.com/tauri-apps/muda/issues/282

### DIFF
--- a/.changes/gtk-feature.md
+++ b/.changes/gtk-feature.md
@@ -1,5 +1,5 @@
 ---
-"muda": major
+"muda": minor
 ---
 
 Make gtk an optional feature (enabled by default)
--- a/.changes/gtk-feature.md
+++ b/.changes/gtk-feature.md
@@ -1,0 +1,5 @@
+---
+"muda": major
+---
+
+Make gtk an optional feature (enabled by default)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,9 @@ categories = ["gui"]
 rust-version = "1.71"
 
 [features]
-default = ["libxdo"]
+default = ["libxdo", "gtk"]
 libxdo = ["dep:libxdo"]
+gtk = ["dep:gtk"]
 common-controls-v6 = []
 serde = ["dep:serde", "dpi/serde"]
 
@@ -42,7 +43,7 @@ features = [
 ]
 
 [target.'cfg(target_os = "linux")'.dependencies]
-gtk = "0.18"
+gtk = { version = "0.18", optional = true }
 libxdo = { version = "0.6.0", optional = true }
 
 [target.'cfg(target_os = "macos")'.dependencies]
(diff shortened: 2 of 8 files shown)


ogoffart:
(Note that this is actually a breaking change for it will break the Linux build if one did `default-features=false`)

Cargo.toml
```diff
@@ -12,8 +12,9 @@ categories = ["gui"]
 rust-version = "1.71"
 
 [features]
-default = ["libxdo"]
-libxdo = ["dep:libxdo"]
+default = ["libxdo", "gtk"]
+libxdo = ["dep:libxdo", "gtk"]
```
amrbashir: i think this can stay as is
```suggestion
libxdo = ["dep:libxdo"]
```

amrbashir (changes requested):
Thank you, could you please add a change file in `.changes` directory? and update the `Cargo features` section in README.md?

.changes/gtk-feature.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": major
```
amrbashir:
```suggestion
"muda": minor
```

This should be a minor since breaking changes are allowed in minor releases for 0.x versions

ogoffart:
Thanks for merging.
Could we get a new release?

amrbashir:
I wanted to make the transition to gtk4 in this release but with my lack of time and effort involved, I guess that is not possible. 

That said, I plan to make a release this week, just need one more PR to merge `IconMenuItem` into `MenuItem` which should be fairly simple but I may skip that as well.

PR #300, opened 2025-06-08
fix(linux): make mnemonic handling consistent across platforms

### LINKED ISSUE #299: Incorrect handling of mnemonics on Linux
There are issues with the implementation of mnemonics for GTK, which, in specific cases, make it impossible to create menus, that behave consistently across all platforms.

1. Escaping of `&` character by doubling it does not work correctly. Two instead of one `&` are displayed.
2. Underscores need to be escaped on Linux in addition to `&` because they too cause mnemonics there.
3. Underscores are incorrectly converted back to ampersands.

## Examples
Original string: `"Test &mnemonic"`
Expected menu item text: Test <ins>m</ins>nemonic
Actual menu item text: Test <ins>m</ins>nemonic
Back conversion: `"Test &mnemonic"`

Original string: `"Test && ampersand && escaping"`
Expected menu item text: Test & ampersand & escaping
Actual menu item text: Test && ampersand && escaping
Back conversion: `"Test && ampersand && escaping"`

Original string: `"Test_underscore"`
Expected menu item text: Test_underscore
Actual menu item text: Test<ins>u</ins>nderscore
Back conversion: `"Test&underscore"`

## Reproduction (uses Tauri)
https://github.com/dgerhardt/tauri-muda-mnemonics-bug

dgerhardt:
`&` and `_` in menu items are now handled on Linux the same way as on Windows.

1. `&&` can be used for escaping and leads to a single displayed `&`.
2. `_` is displayed as is and no longer causes a mnemonic. Escaping is not needed.
3. The back conversion of `_` works correctly.

Fixes #299

### DIFF
--- a/.changes/fix-mnemonics-linux.md
+++ b/.changes/fix-mnemonics-linux.md
@@ -2,10 +2,4 @@
 "muda": patch
 ---
 
-`&` and `_` in menu item labels are now handled on Linux the same way as
-on Windows:
-
-1. `&&` can be used for escaping and leads to a single displayed `&`.
-2. `_` is displayed as is and no longer causes a mnemonic. Escaping is not
-   needed.
-3. The back conversion of `_` works correctly.
+On Linux, fix `&&` resulting in `&&` when it should be just `&`. Also fix `_` not visible and actually adding a mnemonic. This makes the behavior on Linux match the behavior on Windows. 
--- a/.changes/fix-mnemonics-linux.md
+++ b/.changes/fix-mnemonics-linux.md
@@ -1,0 +1,11 @@
+---
+"muda": patch
+---
+
+`&` and `_` in menu item labels are now handled on Linux the same way as
+on Windows:
+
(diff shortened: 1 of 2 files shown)


.changes/fix-mnemonics-linux.md
```diff
@@ -0,0 +1,11 @@
...
+"muda": patch
+---
+
+`&` and `_` in menu item labels are now handled on Linux the same way as
+on Windows:
+
+1. `&&` can be used for escaping and leads to a single displayed `&`.
+2. `_` is displayed as is and no longer causes a mnemonic. Escaping is not
+   needed.
+3. The back conversion of `_` works correctly.
```
amrbashir:
```suggestion
On Linux, fix `&&` resulting in `&&` when it should be just `&`. Also fix `_` not visible and actually adding a mnemonic. This makes the behavior on Linux match the behavior on Windows. 
```

amrbashir (approved):

amrbashir:
thank you

PR #314, opened 2025-08-12
Exclude example files from published package

weiznich:
During our dependency reviews we discovered that the muda crate as uploaded to crates.io includes example data (a png)  that are not used by the rust code. This increases the package size and makes it harder to review the source code as it now includes binary data.

This commit explicitly includes only relevant files. This helps to reduce the package size

Before: 54 files, 482.0KiB (105.0KiB compressed)
After: 39 files, 416.9KiB (87.1KiB compressed)

Overall based on the reduction in the compressed package size and the current download numbers that results to a traffic reduction for crates.io of around 12GB / Month.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/amrbashir/muda"
 documentation = "https://docs.rs/muda"
 categories = ["gui"]
 rust-version = "1.71"
-include = ["README.md", "src/**/*.rs", "Cargo.toml", "LICENSE-APACHE", "LICENSE-MIT"]
+include = ["README.md", "src/**/*.rs", "Cargo.toml", "LICENSE-APACHE", "LICENSE-MIT", "LICENSE.spdx"]
 
 [features]
 default = ["libxdo", "gtk"]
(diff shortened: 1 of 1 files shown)


weiznich:
The CI failure seems to be unrelated to this change due to a new clippy lint: https://github.com/tauri-apps/muda/actions/runs/16910069563/job/47938993462#step:5:1

Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = "https://github.com/amrbashir/muda"
 documentation = "https://docs.rs/muda"
 categories = ["gui"]
 rust-version = "1.71"
+include = ["README.md", "src/**/*.rs", "Cargo.toml", "LICENSE-APACHE", "LICENSE-MIT"]
```
FabianLars:
```suggestion
include = ["README.md", "src/**/*.rs", "Cargo.toml", "LICENSE-APACHE", "LICENSE-MIT", "LICENSE.spdx"]
```
Not really sure if that's required but better safe than sorry when it comes to license stuff 😅

weiznich: Sure, better error on the safe side there.

FabianLars (approved):
thanks :)

PR #321, opened 2025-10-14
Fix padding unconditionally added to top-level submenus for icons even when there's no icon on Linux

signadou:
Currently, `muda` automatically reserves space for icons in top-level submenus; however, because of a bug that I've fixed in this PR, this was being done for all submenus, even if they didn't have icons.  See the screenshots below.

## Screenshots

|Before|After|
|---|---|
|image|image (1)|

### DIFF
--- a/src/platform_impl/gtk/mod.rs
+++ b/src/platform_impl/gtk/mod.rs
@@ -130,8 +130,12 @@ impl Menu {
         return_if_item_not_supported!(item);
 
         for (menu_id, menu_bar) in self.gtk_menubars.iter().filter(|m| *m.0 == id) {
+            let has_icon = matches!(
+                item.kind(),
+                MenuItemKind::Submenu(submenu) if submenu.inner.borrow().icon.is_some()
+            );
             let gtk_item =
-                item.make_gtk_menu_item(*menu_id, self.accel_group.as_ref(), true, true)?;
+                item.make_gtk_menu_item(*menu_id, self.accel_group.as_ref(), true, has_icon)?;
             menu_bar.append(&gtk_item);
             gtk_item.show();
         }
--- a/src/platform_impl/gtk/mod.rs
+++ b/src/platform_impl/gtk/mod.rs
@@ -761,14 +761,32 @@ impl MenuChild {
     pub fn set_icon(&mut self, icon: Option<Icon>) {
         self.icon.clone_from(&icon);
 
-        let pixbuf = icon.map(|i| i.inner.to_pixbuf_scale(16, 16));
+        let pixbuf = icon.as_ref().map(|i| i.inner.to_pixbuf_scale(16, 16));
         for items in self.gtk_menu_items.borrow().values() {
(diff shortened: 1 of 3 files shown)


FabianLars:
thank you for the pr! would you mind adding a small changefile like https://raw.githubusercontent.com/tauri-apps/muda/0ae81ad8b22dd2622e302254ce434d14880087a3/.changes/refactor-simplify-internal-id-usage.md ? :)

signadou:
@FabianLars Added the [changefile](https://github.com/tauri-apps/muda/pull/321/files#diff-0e4f1737e3c4e920bc06d36fb76a6f53045cc8d7367a0ef77ed6e5f0dd0f343a) and fixed the CI on this PR and #322 as well.

src/platform_impl/gtk/mod.rs
```diff
@@ -130,8 +130,12 @@ impl Menu {
         return_if_item_not_supported!(item);
 
         for (menu_id, menu_bar) in self.gtk_menubars.iter().filter(|m| *m.0 == id) {
+            let has_icon = matches!(
+                item.kind(),
+                MenuItemKind::Submenu(submenu) if submenu.inner.borrow().icon.is_some()
+            );
             let gtk_item =
-                item.make_gtk_menu_item(*menu_id, self.accel_group.as_ref(), true, true)?;
+                item.make_gtk_menu_item(*menu_id, self.accel_group.as_ref(), true, has_icon)?;
```
amrbashir: This feels wrong, that last paramter here is `for_menu_bar`, and should not be `has_icon`

The fix should be in implemented in `create_gtk_item_for_submenu`

amrbashir (changes requested):

src/platform_impl/gtk/mod.rs
signadou: @amrbashir I've implemented a better fix in `create_gtk_item_for_submenu` and `create_gtk_item_for_icon_menu_item`.

src/platform_impl/gtk/mod.rs
```diff
@@ -1027,7 +1027,9 @@ impl MenuChild {
             let _ = css_provider.load_from_data(theme.as_bytes());
             style_context.add_provider(&css_provider, gtk::STYLE_PROVIDER_PRIORITY_APPLICATION);
         }
-        box_container.pack_start(&image, false, false, 0);
+        if !for_menu_bar || self.icon.is_some() {
+            box_container.pack_start(&image, false, false, 0);
+        }
```
amrbashir, line 1050: While this works when first creating the menu, this will make a submenu's icon invisible at all times even after calling `Submenu::set_icon` later on, so we also need to modify `MenuChild::set_icon` to add or remove it from box_container

amrbashir (changes requested):

amrbashir (approved):

amrbashir:
Thank you

PR #324, opened 2025-11-20
feat: add FreeBSD platform support

rudrabhoj:
- Add FreeBSD platform implementation
- Update error handling for FreeBSD compatibility
- Modify menu and submenu structures for FreeBSD
- Update Cargo.toml dependencies for FreeBSD build

Basically it was mostly using gtk which works fine with freebsd. Added support for recognizing the freebsd platform and it compiled fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,10 @@ features = [
 gtk = { version = "0.18", optional = true }
 libxdo = { version = "0.6.0", optional = true }
 
+[target.'cfg(target_os = "freebsd")'.dependencies]
+gtk = { version = "0.18", optional = true }
+libxdo = { version = "0.6.0", optional = true }
+
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = "0.6.0"
 objc2-core-foundation = { version = "0.3.0", default-features = false, features = [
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -47,6 +47,10 @@ features = [
 gtk = { version = "0.18", optional = true }
 libxdo = { version = "0.6.0", optional = true }
 
+[target.'cfg(target_os = "freebsd")'.dependencies]
+gtk = { version = "0.18", optional = true }
+libxdo = { version = "0.6.0", optional = true }
```
FabianLars, line 52: you can reuse the linux deps above via the same `any(target_os = "linux", target_os = "freebsd")` you used in code

FabianLars (commented):
thanks for the pr so far! to merge it i'll also need a small changefile (check .changes dir).

lastly, please delete the zellij file :)

Cargo.toml
rudrabhoj: Understood, will make changes and update it!

yaazkal:
any chance this PR gets some tracking again? I'll love to help but don't know how to code in rust :(

My use case is that I'm trying to compile/install openfang and this is one of the dependencies.

jdhg-orbiware:
I was able to compile it in FreeBSD, please check this PR: https://github.com/tauri-apps/muda/pull/331

PR #331, opened 2026-03-15
feat: support FreeBSD Operating System (gtk only)

jdhg-orbiware:
Disclaimer: I don't use rust to code but saw that this change can be pretty straight forward. I do use FreeBSD. This PR works for me to compile muda on FreeBSD.

README file has been updated documenting the support and dependencies needed.

I do appreciate any feedback if there is something wrong with the PR.

Thanks in advance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ features = [
   "Win32_UI_Controls",
 ]
 
-[target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd", target_os = "openbsd"))'.dependencies]
 gtk = { version = "0.18", optional = true }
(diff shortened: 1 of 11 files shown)


delleceste:
[muda-freebsd.patch](https://github.com/user-attachments/files/26243651/muda-freebsd.patch)

This is the patch I used to build muda on FreeBSD in the context of the qbz application 

(https://github.com/vicrodh/qbz)

amrbashir:
Thank you, can you extend this PR to also support other variants of unix systems

```rs
#[cfg(any(
  target_os = "linux",
  target_os = "dragonfly",
  target_os = "freebsd",
  target_os = "netbsd",
  target_os = "openbsd"
))]
```

jdhg-orbiware:
@amrbashir sure, but I was able to test only in FreeBSD. Maybe later?

amrbashir:
no need to test, we can fix them later if anyone has issues but I doubt it, thanks

jdhg-orbiware:
@amrbashir PR updated, thanks for the feedback.

PR #339, opened 2026-04-10
Add `KeyAccelerator` as replacement to `Accelerator` based on `keyboard_types::Key`

### LINKED ISSUE #294: Cannot use "+" as an accelerator modifier
`"PLUS" | "+" => Ok(Plus),`

is missing from [accelerators.rs](https://github.com/tauri-apps/muda/blob/dev/src/accelerator.rs#L255)

### LINKED ISSUE #333: Allow accelerators based on keyboard-types `Key`
Hi,
First of all, thank you all for developing and maintaining muda. We are happily using it in Slint for our native menu bars and context menus. 😊 

Recently we ran into the limitation that muda only allows creating accelerators based on keyboard-types `Code` type, which doesn't allow expressing all keyboard shortcuts that Slint supports.
Concretely, the `Code` type is limited to physical key codes, even though accelerators usually use "logical" keys - e.g. the string that was produced by the keypress, not the physical key position.

That means that Muda currently does not support keyboard shortcuts like `Control + +`/`Control + {`/`Control + €`, etc.

**I would like to contribute to muda to add support for these kinds of keyboard shortcuts.**

In my opinion the right approach is to add support for the `Key` type from keyboard-types instead of the `Code` type.
`Key` is the right type because it represents a logical keypress that produced a certain string instead of a physical key position.

Ideally, I think accelerators should entirely migrate to using the `Key` type, but I can understand that for backwards compatibility it is probably better to continue offering the constructor that accepts a `Code`.

Would you consider merging a pull request that adds an additional constructor for accelerators that use the `Key` type?
Also, would you prefer keeping the existing `Code` constructor, or remove it entirely?

LeonMatthes:
Add a new `KeyAccelerator` struct based on `keyboard_types::Key` alongside
the existing `Accelerator` (based on `Code`). This enables expressing
logical key shortcuts like Ctrl++, Ctrl+€ that physical key codes
cannot represent (see https://github.com/tauri-apps/muda/issues/333).

As this is the more modern approach to handling shortcuts in menu items,
we can internally deprecate the Accelerator struct and move everything
to use the KeyAccelerator struct.

The existing public Accelerator API is 100% unchanged though, making this
a non-breaking additive change from the outside.

Closes #333 
Closes #294


src/platform_impl/macos/accelerator.rs
```diff
@@ -2,114 +2,62 @@
...
 
-use keyboard_types::{Code, Modifiers};
+use keyboard_types::{Key, Modifiers};
 use objc2_app_kit::NSEventModifierFlags;
 
-use crate::accelerator::{Accelerator, AcceleratorParseError};
+use crate::accelerator::{AcceleratorParseError, KeyAccelerator};
 
-impl Accelerator {
+impl KeyAccelerator {
```
LeonMatthes, line 10: Note: This is technically a breaking change.
However, this is really platform API.
Do we care about this as a breaking change?

amrbashir: I think those were accidental public APIs, and I am fine with just hiding these functions behind `pub(crate)` instead

LeonMatthes: done

examples/wry.rs
```diff
@@ -111,11 +111,11 @@ fn main() -> wry::Result<()> {
         .append_items(&[&file_m, &edit_m, &window_m])
         .unwrap();
 
-    let custom_i_1 = MenuItem::new(
-        "C&ustom 1",
-        true,
-        Some(Accelerator::new(Some(Modifiers::ALT), Code::KeyC)),
-    );
+    let custom_i_1 = MenuItem::new("C&ustom 1", true, None);
+    custom_i_1.set_key_accelerator(Some(muda::accelerator::KeyAccelerator::new(
```
amrbashir: let's do `use muda::accelerator::KeyAccelerator` at top instead

src/builders/check.rs
```diff
@@ -46,23 +49,46 @@ impl CheckMenuItemBuilder {
     }
 
     /// Set this check menu item accelerator.
+    ///
+    /// (Note that setting an accelerator will override any existing key_accelerator)
```
amrbashir:
```suggestion
    /// (Note that setting an accelerator will override any existing [.key_accelerator()](Self::key_accelerator))
```
and similarly for other menu items

src/builders/check.rs
```diff
@@ -46,23 +49,46 @@ impl CheckMenuItemBuilder {
...
+        self.key_accelerator = accelerator
+            .map(|a| a.try_into())
+            .transpose()?
+            .map(KeyAccelerator::from);
+        Ok(self)
+    }
+
+    /// Set this check menu item accelerator using a [`KeyAccelerator`].
+    ///
+    /// (Note that setting a key_accelerator will override any existing accelerator)
```
amrbashir:
```suggestion
    /// (Note that setting a key_accelerator will override any existing [.accelerator()](Self::accelerator))
```

and similarly for other menu items

src/items/check.rs
```diff
@@ -109,8 +113,22 @@ impl CheckMenuItem {
     }
 
     /// Set this check menu item accelerator.
+    ///
+    /// (Note that setting an accelerator will override any existing key_accelerator)
```
amrbashir:
```suggestion
    /// (Note that setting an accelerator will override any existing [.set_key_accelerator()](Self::set_key_accelerator))
```
and similarly for other menu items

src/items/check.rs
```diff
@@ -109,8 +113,22 @@ impl CheckMenuItem {
...
     pub fn set_accelerator(&self, accelerator: Option<Accelerator>) -> crate::Result<()> {
-        self.inner.borrow_mut().set_accelerator(accelerator)
+        self.inner
+            .borrow_mut()
+            .set_key_accelerator(accelerator.map(KeyAccelerator::from))
+    }
+
+    /// Set this check menu item accelerator using a [`KeyAccelerator`].
+    ///
+    /// (Note that setting a key_accelerator will override any existing accelerator)
```
amrbashir:
```suggestion
    /// (Note that setting a key_accelerator will override any existing [.set_accelerator()](Self::set_accelerator))
```
and similarly for other menu items

src/accelerator.rs
```diff
@@ -4,38 +4,40 @@
...
 //!
 //! # Examples
 //! They can be created directly
 //! ```no_run
-//! # use muda::accelerator::{Accelerator, Modifiers, Code};
-//! let accelerator = Accelerator::new(Some(Modifiers::SHIFT), Code::KeyQ);
-//! let accelerator_without_mods = Accelerator::new(None, Code::KeyQ);
+//! # use muda::accelerator::{KeyAccelerator, Modifiers, Key};
+//! let accelerator = KeyAccelerator::new(Some(Modifiers::SHIFT), Key::Character("q".to_owned()));
+//! let accelerator_without_mods = KeyAccelerator::new(None, Key::Character("q".to_owned()));
```
amrbashir: we could keep both variants in module documentation for now.

src/accelerator.rs
```diff
@@ -48,6 +50,10 @@ pub enum AcceleratorParseError {
 /// A keyboard shortcut that consists of an optional combination
 /// of modifier keys (provided by [`Modifiers`] and
 /// one key ([`Code`]).
+///
+/// ⚠️This struct cannot represent all shortcuts found on non-U.S. keyboard layouts and might
+/// be deprecated in the future.
+/// Please use [`KeyAccelerator`] instead.
```
amrbashir:
```suggestion
/// ## Warning
///
/// This struct cannot represent all shortcuts found on non-U.S. keyboard layouts and might be deprecated in the future.
/// Please use [`KeyAccelerator`] instead.
```

src/accelerator.rs
```diff
@@ -321,10 +252,360 @@ fn parse_key(key: &str) -> Result<Code, AcceleratorParseError> {
...
+
+fn split_key_and_modifiers(accelerator: &str) -> Result<(&str, &str), AcceleratorParseError> {
+    let accelerator = accelerator.trim();
+    if accelerator.is_empty() {
+        return Err(AcceleratorParseError::InvalidFormat(String::new()));
+    }
+
+    // Separate modifier part from key part using rfind('+').
+    // This correctly handles '+' as the key: "Ctrl++" → rfind gives the last '+',
+    // leaving "Ctrl+" as the modifier part and "" as raw key → key is '+'.
```
amrbashir:
```suggestion
    // Separate modifier part from key part using rfind('+').
    // This correctly handles '+' as the key: "Ctrl++" -> rfind gives the last '+',
    // leaving "Ctrl+" as the modifier part and "" as raw key -> key is '+'.
```

src/platform_impl/macos/accelerator.rs
```diff
@@ -2,114 +2,62 @@
...
-            Code::Digit4 => "4".into(),
-            Code::Digit5 => "5".into(),
-            Code::Digit6 => "6".into(),
-            Code::Digit7 => "7".into(),
-            Code::Digit8 => "8".into(),
-            Code::Digit9 => "9".into(),
-            Code::Comma => ",".into(),
-            Code::Minus => "-".into(),
-            Code::Period => ".".into(),
-            Code::Space => "\u{0020}".into(),
```
amrbashir, line 55: I think this still needs to be special handled, which was not done in the new diff

src/platform_impl/gtk/accelerator.rs
```diff
@@ -112,72 +77,54 @@ fn modifiers_to_gdk_modifier_type(modifiers: Modifiers) -> gdk::ModifierType {
...
-
-        Code::ControlLeft => Control_L,
-        Code::AltLeft => Alt_L,
-        Code::ShiftLeft => Shift_L,
-        Code::MetaLeft => Super_L,
-
-        Code::ControlRight => Control_R,
-        Code::AltRight => Alt_R,
-        Code::ShiftRight => Shift_R,
-        Code::MetaRight => Super_R,
```
amrbashir, line 132: these were removed, I don't remember if they were ever used, but let's bring it back just to be sure

amrbashir (changes requested):
Than you, this looks fantastic. Just left a few nitpicks.

src/platform_impl/macos/accelerator.rs
LeonMatthes: The Unicode codepoint 0020 is just the space character (`" "`), so this doesn't need special handling. (Caught me by surprise as well recently) :sweat_smile: 
Key just represents this as `Key::Character(" ")`, which is the same as `Key::Character("\u{u0020}")` 
https://unicode-explorer.com/c/0020

Somewhat related fun-fact, you can use this to achieve multi-line indented string-literals in Rust:

e.g. we do this in Slint here:
```rust
// \x20 equals to a space (needed to avoid the trailing \ eating
// the indentation)
&format!(
    "{text} is not a cross-platform modifier\n\
    Use cross-platform modifier names instead:\n\
    \x20   ⌘ command -> Control\n\
    \x20   ⌥ option -> Alt\n\
    \x20   ^ control -> Meta\n\
    \x20   ⇧ shift -> Shift"
),
```

src/platform_impl/gtk/accelerator.rs
LeonMatthes: Ah, so these are the few keys that are different between Code and Key.
These don't have a native representation in the key type because they're not logical keys, they are physical ones. We could, for example, represent them in the key character with a special Unicode sequence. This is what we do in Slint and it works pretty well.

However, I just tested them and even though the shortcut shows up in the menu item under GDK,  I can't get the shortcut to activate, so I think this is broken at the moment anyhow.

I'm in favor of removing them and we'll see if anyone notices.
I doubt anyone will because I don't think this was working in the first place.

LeonMatthes:
> Than you, this looks fantastic. Just left a few nitpicks.

thank you for the quick responses and review :raised_hands:

LeonMatthes:
@amrbashir The CI failures should be fixed now, can you re-approve the workflow? :crossed_fingers:

src/platform_impl/macos/accelerator.rs
amrbashir: Yeah it was weird when I first saw that but at the time of implementation I didn't have a mac to test on, so I just accepted it as is.

I just tested to be 100% sure, works fine.

src/platform_impl/gtk/accelerator.rs
amrbashir: yeah I looked at other platforms and I think we can safely remove. It was only implemented in GTK, so I doubt any would is using it.

amrbashir:
Looks like CI is still broken

LeonMatthes:
@amrbashir Sorry about that, I forgot to save some files from the editor :facepalm: 
I ran this patch through Slint CI now and it passed, so should at least build on macOS and windows now. Hopefully clippy doesn't have anything to complain about :crossed_fingers:

amrbashir (approved):
Thank you

amrbashir (commented):
apologies, I forgot to mention we need a change file `.changes` directory (with `minor` bump), similar to https://github.com/tauri-apps/muda/blob/dev/.changes/fix-subclassing-bug.md

LeonMatthes:
@amrbashir change file added

amrbashir:
Thank you

PR #352, opened 2026-05-05
fix(macos): clamp context-menu popup so it fits on-screen

vdavid:
### Problem

`popUpMenuPositioningItem:atLocation:inView:` does not auto-flip or clamp the menu when it would overflow the screen, so right-clicking near the bottom edge of a window produces a menu with scroll arrows instead of one that's fully visible. It looks bad.
Easy repro: open any consumer of `ContextMenu::show_context_menu_for_nsview`, right-click within ~one menu-height of the bottom of the screen → the menu opens with a downward scroll arrow rather than flipping upward like a standard AppKit contextual menu. Not nice.

### Fix

Switch `show_context_menu` to `+[NSMenu popUpContextMenu:withEvent:forView:]`, which is AppKit's idiomatic contextual-menu entry point and handles auto-flip, screen clamp, menu bar / Dock margins, and multi-monitor positioning natively. Driving it requires an `NSEvent`, so we synthesize a right-mouse-down event with `+[NSEvent mouseEventWithType:location:modifierFlags:timestamp:windowNumber:context:eventNumber:clickCount:pressure:]` whose `locationInWindow` is the popup point in the target view's window's coordinate space.

`popUpContextMenu:withEvent:forView:` requires a non-nil `NSView`. The current API already takes a `*const c_void` view that the caller guarantees is valid, so we pass it through directly. If `+mouseEventWithType:` ever fails to synthesize an event, the code falls back to `popUpMenuPositioningItem:atLocation:inView:` so the menu still appears.

### Edge cases handled

- The popup location is computed in the view's window's coordinate space (the same window the synthesized event targets via `[window windowNumber]`), which is what `popUpContextMenu:withEvent:forView:` expects.
- Bottom edge, right edge, menu-bar / Dock margins, and multi-monitor: handled by AppKit's contextual-menu code path. Future macOS improvements to that code path apply automatically.

### Behavior change risks

User-visible end result is the same: menus that previously fit on-screen still fit, menus that previously overflowed now flip up / shift inward instead of showing scroll arrows. Internally the implementation now matches Apple's idiom, so future macOS tweaks to contextual-menu positioning come for free.

### Cargo.toml

Adds `"NSGraphicsContext"` to `objc2-app-kit` features (required by `+mouseEventWithType:...`, even though we pass `nil` for the context argument) and `"NSDate"` to `objc2-foundation` features (for the timestamp). No `NSScreen` feature needed — AppKit handles screen geometry internally.

### Test plan

I did this:

- Ran `cargo build`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, they all pass.
- Manually tested in [Cmdr](https://getcmdr.com) (my file manager): right-clicked near the bottom and right edges of the screen; now it produces a fully visible, properly clamped menu instead of a scrolling one. I did not multi-monitor verify, unfortunately, because I have no secondary screen now. :/

### Changelog

Covector entry included as `.changes/macos-popup-clamp.md` (`patch` bump).

### DIFF
--- a/.changes/macos-popup-clamp.md
+++ b/.changes/macos-popup-clamp.md
@@ -2,4 +2,4 @@
 "muda": patch
 ---
 
-On macOS, context-menu popups no longer overflow off the bottom or right edge of the screen and fall back to scroll arrows. `show_context_menu_for_nsview` now nudges the popup anchor so the menu fits inside the active screen's visible frame, which accounts for the menu bar, the Dock, and multi-monitor setups.
+On macOS, context-menu popups are clamped to avoid overflowing off the bottom or right edge of the screen.
--- a/.changes/macos-popup-clamp.md
+++ b/.changes/macos-popup-clamp.md
@@ -1,0 +1,5 @@
+---
+"muda": patch
+---
+
+On macOS, context-menu popups no longer overflow off the bottom or right edge of the screen and fall back to scroll arrows. `show_context_menu_for_nsview` now nudges the popup anchor so the menu fits inside the active screen's visible frame, which accounts for the menu bar, the Dock, and multi-monitor setups.
--- a/src/platform_impl/macos/mod.rs
+++ b/src/platform_impl/macos/mod.rs
@@ -25,10 +25,11 @@ use objc2_app_kit::{
     NSAboutPanelOptionApplicationVersion, NSAboutPanelOptionCredits, NSAboutPanelOptionVersion,
     NSApplication, NSControlStateValueOff, NSControlStateValueOn, NSEvent, NSEventModifierFlags,
     NSImage, NSImageName, NSMenu, NSMenuDelegate, NSMenuItem, NSRunningApplication, NSView,
+    NSWindow,
 };
 use objc2_foundation::{
     ns_string, MainThreadMarker, NSAttributedString, NSDictionary, NSInteger, NSObject, NSPoint,
-    NSSize, NSString,
+    NSRect, NSSize, NSString,
 };
 
 use self::util::strip_mnemonic;
@@ -1315,6 +1316,9 @@ impl NativeIcon {
     }
 }
 
+/// How far off the screen edge a nudged-up menu is kept, in points.
+const SCREEN_EDGE_MARGIN: f64 = 4.0;
+
 unsafe fn show_context_menu(
     ns_menu: &NSMenu,
     view: *const c_void,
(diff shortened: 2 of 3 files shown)


vdavid:
Hi, @amrbashir! Any chance I could get a review on this one? Sorry for not pinging you earlier, I somehow just assumed someone would review the PR.

amrbashir:
Could you provide a few screenshots, aka before and after? Also do you know if Windows and Linux handle the clamp automatically or do we need to also implement it there as well

vdavid:
Sure!
Here is a before/after from the repo's `tao` example.

Setup was: window at the bottom of my screen, 12 extra items added so the menu is taller than the gap below the click point. Then I always right-clicked roughly where you see the top-left of the context menu in the first screenshot below.

## Before (`popUpMenuPositioningItem:atLocation:inView:`)

You see, there is all that space above, and that sad context menu is just sitting down there, with the downward scroll arrow at its bottom:

muda-before-my-pr

(Btw the old path is broken this same way in both the "explicit position" and the "position-less/mouse-location" modes.)

## After (`popUpContextMenu:withEvent:forView:`)

Much nicer, BUT with an unwanted extra item ("AutoFill"):

muda-after-my-pr

(Fixed, see the scrshot below)

## After (with `setAllowsContextMenuPlugIns(false)`)

While taking the screenshots, I caught an unintended side effect of my change: `popUpContextMenu:` lets macOS append its own items to the menu and a divider and "AutoFill" showed up as a result. The legacy call never did that, so I pushed a commit that opts out via `setAllowsContextMenuPlugIns(false)`. Menu contents stay exactly what the consumer built.

muda-after-my-pr-then-fixed

## Windows and Linux

On Windows and Linux, **no work needed**. Both clamp natively on muda's existing calls:
- Windows: muda calls `TrackPopupMenu`, and the OS repositions popup menus to fit the work area on its own. (I don't have a Windows box to demo it, sorry.)
- Linux: muda calls `gtk_menu_popup_at_rect`, and `GtkMenu:anchor-hints` defaults to `GDK_ANCHOR_FLIP_X | FLIP_Y | SLIDE_X | SLIDE_Y | RESIZE_X | RESIZE_Y`, so GTK flips/slides the menu to keep it on the monitor: https://docs.gtk.org/gtk3/property.Menu.anchor-hints.html

Only macOS was broken: `popUpMenuPositioningItem:` positions manually, without the OS's clamping treatment. And my changes only affect macOS.

WDYT, good to merge?

vdavid:
@amrbashir, sorry, forgot to mention you, not sure how your notifications are set up, so just wanted to send a polite ping.

vdavid:
@amrbashir, any chance to get this PR merged? I've kept it fresh with the updates. I'd love to have this in to fix popup heights.

amrbashir:
no matter what I tried to reproduce using `tao.rs` example, it worked fine and clamped correctly inside the monitor. I am on macOS 26. Can you share your repro and system information?

vdavid:
1-min version:

→ [Quick screen recording link](https://www.loom.com/share/1f3be33120cc4b539e95fe77d83e38ee) ←

More:

Sure, here's everything.

## My system info

- MacBook Pro (`Mac15,9`), M3 Max, macOS 26.5.2 (`25F84`)
- Built-in Liquid Retina XDR, scaled so it "looks like" 3360×1890, 2x backing scale
- Dock at the bottom, always visible, tile size 48

## Why I think you couldn't repro

I think it's the menu height. `file_m` in `examples/tao.rs` has only six items (~130 px tall), and the Dock eats the bottom ~80 px, so you pretty much can't get the click point close enough to the screen edge to make it overflow. The bug only shows when the menu is taller than the space below the click.

My original repro added 12 items by patching the example. But actually, you don't need to patch anything: clicking "Custom 1" inserts a new item into `file_m`, so the example can grow its own menu.

## This 1-min repro works

1. `cargo run --example tao`
2. Right-click in the window, click "Custom 1". That inserts a "New Menu Item" into `file_m`. Repeat ~12 times, until the context menu is ~500 px tall.
3. Drag the window down so its bottom edge sits right above the Dock.
4. Right-click ~30 px above the window's bottom edge.

On `dev` I get the menu anchored at the click point with a downward scroll arrow at the bottom, and a big pile of unused space above it. (Still reproduces on dev as of today, just checked.) With this PR it flips up and shows in full.

[Also a screen recording so you don't even need to run the example](https://www.loom.com/share/1f3be33120cc4b539e95fe77d83e38ee)

src/platform_impl/macos/mod.rs
```diff
@@ -1325,19 +1326,60 @@ unsafe fn show_context_menu(
...
 
-    ns_menu.popUpMenuPositioningItem_atLocation_inView(None, location, in_view)
+    // `popUpContextMenu:withEvent:forView:` lets macOS append contextual-menu
+    // plug-in items (AutoFill, Services, third-party plug-ins) to the menu.
+    // The legacy positioning call never did that, so opt out to keep the
+    // menu's contents exactly what the consumer built.
+    ns_menu.setAllowsContextMenuPlugIns(false);
+
+    NSMenu::popUpContextMenu_withEvent_forView(ns_menu, &event, view);
+    true
```
amrbashir: this is a regression, I think we can do the clamping ourselves and keep using the old API so that we know if a user selected or cancelled the menu.

Electron does something similar here https://github.com/electron/electron/blob/fe4cffac230a21114e645834eab06072f4ea2034/shell/browser/api/electron_api_menu_mac.mm#L121

vdavid: Good catch, you're right. Rewrote it the way you suggested.

vdavid:
Done, rewrote it to fit the anchor ourselves, like the Electron code.
With that, the `setAllowsContextMenuPlugIns` commit was no longer needed, so the PR is now down to one commit. The math is a pure function now, with unit tests for the six cases.

QA-ed it, it looks very nice after the changes, both at the bottom of the screen and at the very right of it.

amrbashir (approved):

.changes/macos-popup-clamp.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": patch
+---
+
+On macOS, context-menu popups no longer overflow off the bottom or right edge of the screen and fall back to scroll arrows. `show_context_menu_for_nsview` now nudges the popup anchor so the menu fits inside the active screen's visible frame, which accounts for the menu bar, the Dock, and multi-monitor setups.
```
amrbashir:
```suggestion
On macOS, context-menu popups are clamped to avoid overflowing off the bottom or right edge of the screen.
```

amrbashir (approved):

amrbashir:
Thank you

PR #353, opened 2026-05-05
feat(macos): add set_text_with_secondary for two-color menu labels

vdavid:
## Problem

`NSMenuItem` supports `setAttributedTitle:`, which lets AppKit render menu labels with mixed styling. For example, Finder uses it to draw entries like `Preview (default)` where the " (default)" part is dimmed via `NSColor.secondaryLabelColor`, which looks pretty cool.
Muda doesn't currently expose this, so consumers either ship plain text or reach into AppKit themselves via `objc2`/FFI to grab the `NSMenuItem` and set the attributed title manually. That's painful and bypasses muda's `text` bookkeeping.

Why this is important to me: I want the same styling for the "Open with" submenu in [Cmdr](https://getcmdr.com) (my macOS file manager) that Finder has. Right now I'm running a sentinel-based workaround in a local muda fork to get the same effect, but it's very ugly. This PR (along with a future Tauri PR) is the proper typed API to replace it, and it seems helpful to other consumers, too.

## Fix

In this PR I add a typed `set_text_with_secondary(primary: &str, secondary: Option<&str>)` method on `MenuItem` and `IconMenuItem`.
On macOS, it builds an `NSMutableAttributedString` with two runs at the standard menu font (`NSFont.menuFontOfSize(0.0)`), applies `NSColor.secondaryLabelColor` to the secondary range, and calls `setAttributedTitle:` on each underlying `NSMenuItem`. Passing `None` for `secondary` clears the attributed title and falls back to plain `setTitle:`.

The `(primary, secondary)` pair is stored on `MenuChild`, so the attributed title is applied both immediately to existing native items AND at lazy creation time when the muda item is later appended to a parent submenu. Without this, calling `set_text_with_secondary` before appending would silently produce plain text because the freshly-created `NSMenuItem` wouldn't know it was supposed to have an attributed title.

Note: I considered exposing a more general "set arbitrary attributed string" API but kept the surface small to match a concrete HIG pattern. Happy to broaden it (or extend to `CheckMenuItem`/`Submenu`) if you (the maintainers) prefer.

## Edge cases

- `secondary == None` clears `attributedTitle` so the item goes back to plain styling.
- The method also updates the cached `text` field to the concatenation of `primary` and `secondary`, so `text()` returns what's visible (matching how `setAttributedTitle:` actually behaves at draw time).
- The mnemonic-stripping logic (`strip_mnemonic`) is applied to both parts, consistent with `set_text`.

## Behavior change risks

Nothing, really.

- `set_text_with_secondary` is a pure addition.
- I tweaked `set_text` to also call `setAttributedTitle: nil` on existing native items. This is a no-op for items that never had an attributed title set, but ensures `set_text` cleanly reverts styling after `set_text_with_secondary`.

## Cargo.toml

The macOS impl now uses `NSColor`, `NSFont`, `NSForegroundColorAttributeName`, `NSFontAttributeName`, `NSMutableAttributedString`, and `NSRange`, so I enabled the matching features on `objc2-app-kit` (`NSAttributedString`, `NSColor`, `NSFont`) and `objc2-foundation` (`NSRange`).
Also pulled in the `AnyThread` trait from `objc2` for `NSMutableAttributedString::alloc()`.

## Test plan

I've done this:

- Ran `cargo build`, `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, they all pass on macOS.
- Added unit tests to check that `text()` returns the concatenation after `set_text_with_secondary`, and `attributed_title_parts_round_trip` confirms the stored `(primary, secondary)` pair is populated by `set_text_with_secondary`, cleared by `set_text_with_secondary(.., None)`, and cleared by `set_text` (so plain `set_text` is a clean revert).
- Manually verified by temporarily adding two demo items to `examples/tao.rs` (a `MenuItem` and an `IconMenuItem`, both calling `set_text_with_secondary("…", Some(" (…)"))`), running `cargo run --example tao`, and confirming the suffix renders dimmed in `NSColor.secondaryLabelColor` matching Finder's "Open with → App Name (default)" exactly. Example change kept local.

## Changelog

`.changes/menu-item-secondary-label.md` (minor bump, since it's a public API addition).

**Note:** Tauri's `tauri::menu::IconMenuItem` doesn't currently expose enough surface to do this through its wrapper, so I'm planning a follow-up Tauri PR to delegate to this method. I'm filing the muda piece first since it's the prerequisite.


vdavid:
Hi, @amrbashir! Any chance I could get a review on this one? Sorry for not pinging you earlier, I somehow just assumed someone would review the PR.

amrbashir:
I feel like there maybe needs for more APIs similar to this, and to avoid adding each one, we should instead expose the native item so users can call the APIs they need directly. Maybe something like:
```rs

let menu_item = MenuItem::new("item", true, None);
menu_item.with_native_item(|item| item.setAttributedTitle());
```

vdavid:
I like the idea to just expose the native item, but I think it wouldn't work:
- muda creates native items lazily: one `NSMenuItem` per menu the item is attached to (the `ns_menu_items` map).
- A run-once closure only styles the native items that exist at call time.
- If you append the muda item to another menu later, the fresh `NSMenuItem` comes out unstyled.

That's exactly why this PR stores the `(primary, secondary)` pair on `MenuChild` and re-applies it at native-item creation time.

So a functioning `with_native_item` would need to store the closure and replay it on every future native-item creation. Which is doable, but a bigger design than it looks:

- needs closure storage + `Fn` bounds (it runs multiple times, possibly long after the call)
- per-platform signature (`&NSMenuItem` on macOS, `HMENU` + id on Windows, N `GtkMenuItem`s per child on Linux)
- consumers can silently break muda's bookkeeping (for example, `setTitle:` behind muda's back makes `text()` lie)

There's also my actual end goal: using this through Tauri's menu wrapper. A typed setter is easy to delegate through `tauri::menu`; a closure-over-`objc2`-types API is much harder to surface there.

So my take: I'd merge the typed setter now. It follows muda's existing pattern of platform-specific typed APIs (`set_native_icon` is already macOS-only), and it matches a concrete HIG pattern (Finder-style dimmed suffixes). If more "expose this AppKit thing" requests come in later, that's the right time to design the stored-and-replayed solution and this setter can then become a thin wrapper over it.

That said, if you'd rather have the stored-and-replayed solution, I can build it (in this PR or a follow-up) with the setter on top. I just think it's a bit premature while this is the only use case.

What do you think?

vdavid:
@amrbashir, sorry, forgot to mention you, not sure how your notifications are set up, so just wanted to send a polite ping.

vdavid:
@amrbashir, any chance to get this PR merged? I've kept it fresh with the updates. I'd love to use standard `muda` instead of maintaining my fork, and this feels like a genuinely nice addition for all consumers.

amrbashir:
I will be checking these PRs soon, thanks for your patience

amrbashir:
I still don't like the idea of adding this API, and I think we should find a better generic way so that we don't have to keep adding similar niche APIs and I don't have a good API design atm.

vdavid:
Yeah, I agree that this is a tricky design decision. I've brought an idea for a more generic version.

How about if, instead of exposing the native item, we keep the API **pure data** and let muda keep owning the text:

```rust
#[non_exhaustive]
pub enum TextStyle {
    Normal,
    Secondary,
}

item.set_styled_text([
    ("Preview", TextStyle::Normal),
    (" (default)", TextStyle::Secondary),
]);
```

So it's a list of runs with semantic styles, instead of the `set_text_with_secondary(primary, secondary)` that this PR had before. It's the same idea, but more general.

This seems to be better than exposing the native item, because

- This leaves **nothing to break**. muda still owns the string, so `text()` stays truthful (it's just the concat of the runs).
- It **works for all OSes**. The API surface is platform-neutral even though only macOS implements real styling for now.
- **Pure data crosses the Tauri wrapper trivially.** A `Vec<(String, TextStyle)>` is a two-line pass-through in `tauri::menu`. A closure over `objc2` types isn't, and it would drag `objc2` into Tauri's public API too (plus pin muda's semver to `objc2`'s). This matters a lot to me, since using this from Tauri is my end goal.
- The enum is `#[non_exhaustive]`, so adding a style later isn't a breaking change. The enum grows, the API surface doesn't.

## Other backends

Just plain text fallback for now:

- **Windows**: `TrackPopupMenu` takes a plain string, so real styling needs owner-draw, I'd not go into that for now.
- **GTK3**: doable later via Pango markup on the `AccelLabel` (`<span alpha="...">` or the dim-label style).
- **GTK4** (if I understand correctly): menus come from `gio::Menu` + `PopoverMenu`, and per-item markup isn't really available there AFAICT. It probably means plain text for good.

## My proposed implementation

I've actually updated the PR with my suggested shape. Not a lot changed, just generalized:

- Now storing the runs on `MenuChild`.
- Applying to the existing `NSMenuItem`s right away, and re-apply in `create_ns_item` (this is the bit that matters: muda creates one `NSMenuItem` per menu the item is attached to, lazily, so anything applied only at call time gets lost when the item is later appended somewhere else).
- An all-`Normal` label stores nothing and takes the plain `setTitle:` path, so you don't pay for an attributed title that draws identically, and no stale one is left behind on items created later.
- `set_text` now clears the runs, so it's a clean revert.
- `text()` returns the concatenation, which matches what's actually drawn.

If you don't like this one either, pls tell me more about your concerns and we can work it out.

src/items/icon.rs
```diff
@@ -183,6 +183,46 @@ impl IconMenuItem {
         self.inner.borrow_mut().set_text(text.as_ref())
     }
 
+    /// Set the item's label as a sequence of styled runs, so one part of the label can be
```
amrbashir: I don't like the word "run" here or in the rest of API, I think it is unclear. What about just "as a sequence of styled text"

src/items/icon.rs
```diff
@@ -183,6 +183,46 @@ impl IconMenuItem {
...
+    /// ```
+    ///
+    /// [`IconMenuItem::text`] returns the concatenation of the runs, which is what the item
+    /// actually draws. [`IconMenuItem::set_text`] clears the styling.
+    ///
+    /// ## Platform-specific:
+    ///
+    /// - **Windows / Linux**: the runs are concatenated and set as plain text; styles are
+    ///   not visually distinguished.
+    pub fn set_styled_text<S: AsRef<str>>(&self, runs: impl IntoIterator<Item = (S, TextStyle)>) {
```
amrbashir: we need to add similar method on `MenuItemBuilder`, `IconMenuItemBuilder`..etc so users can create it from the getgo, instead of creating with a text and later changing it.

amrbashir: It seems like you only added this method to normal and icon menu items, what about `CheckMenuItem` and `Submenu` ?

amrbashir (commented):
Thank you, this new design is something I can agree to.

src/items/icon.rs
vdavid: Agreed. Rephrased to "sequence of" / "part" everywhere.

src/items/icon.rs
vdavid: Good catch, sorry, both of those have it now, too.

vdavid: Sorry for missing those. Added, all four builders have `.styled_text(...)` now. It overrides `.text()` when both are set.

vdavid:
Thanks for the super fast review! I've fixed all three points. I've also moved the platform split into one `apply_styled_text` helper, so each wrapper is a single line. Tests cover all four item types and all four builders. Rebased on today's `dev` and green, intended to be mergeable as-is.

amrbashir (approved):

PR #357, opened 2026-05-19
fix(windows): prevent dark menubar item text truncation

mattico:
The dark menubar draw path fetched menu item text with an incorrectly sized UTF-16 buffer, which truncated top-level labels during custom drawing in dark mode. Previously it used `size_of_val(&label)`, which returned the size of the Vec struct itself rather than its contents and limited retrieval to about 10 UTF-16 code units on 64-bit Windows.

Use the standard two-step GetMenuItemInfoW pattern: first query the required string length, then allocate a buffer large enough for the full label plus the trailing null terminator before fetching the text.

Also returns early if either Win32 call fails so the draw path skips the label instead of using incomplete data.

**Before:**
image

**After:**
image

### DIFF
--- a/.changes/fix-win-dark-menubar.md
+++ b/.changes/fix-win-dark-menubar.md
@@ -1,0 +1,5 @@
+---
+"muda": patch
+---
+
+On Windows, fixed a bug that would truncate menubar items text to ~10 characters in dark-mode.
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
Thank you, can you please add a change file in `.changes` directory?

mattico:
Done!

amrbashir (approved):
Thank you

PR #361, opened 2026-06-04
fix(macos): replace raw MenuChild pointer with Rc

### LINKED ISSUE #328: macOS: use-after-free crash when calling set_menu while menu is displayed
## Summary

Calling `set_menu()` on a tray icon (via `tray-icon`) while the menu is currently displayed causes a use-after-free. If the user then clicks a menu item, the app crashes inside `fire_menu_item_click` with a panic in `to_png()` (`ZeroWidth`) or a segfault — depending on what the freed memory looks like.

## Root cause

`MudaMenuItem` stores a raw `*const MenuChild` pointer as an ivar (mod.rs:947):

```rust
// FIXME: Use `Rc` or something else to access the MenuChild.
#[ivars = Cell<*const MenuChild>]
struct MenuItem;
```

When `set_menu(new_menu)` is called:
1. The old `Box<dyn ContextMenu>` is dropped → `MenuChild` items are freed
2. But old `NSMenuItem`s survive (retained by the still-visible `NSMenu`)
3. User clicks a stale menu item → `fire_menu_item_click` dereferences the freed pointer
4. Use-after-free → reads garbage → crash

## Crash we hit

```
PANIC on thread 'main' at muda-0.17.1/src/platform_impl/macos/icon.rs:34:53:
called `Result::unwrap()` on an `Err` value: Format(FormatError { inner: ZeroWidth })
```

The freed memory was reinterpreted as a `MenuChild` with an `Icon` containing `width=0`. `to_png()` encodes it → png crate rejects zero width → `.unwrap()` panics. Since this runs inside an `extern "C"` Obj-C callback (`fireMenuItemAction:`), the panic crosses a nounwind boundary → immediate abort. `catch_unwind` cannot help.

## Secondary issue: `to_png()` uses `.unwrap()`

Even without the use-after-free, `PlatformIcon::to_png()` (icon.rs:34) should return `Result` instead of using `.unwrap()`. The tray-icon crate's version of the same function already does this. And `Icon::from_rgba(vec![], 0, 0)` currently succeeds (the validation allows zero dimensions), so a zero-dimension Icon can be constructed legally.

## Reproduction

This triggers in any app that periodically calls `tray.set_menu(Some(new_menu))` to update the tray menu while the user has the menu open. Related to #128, #202, #233 — all are manifestations of the dangling `*const MenuChild`.

## Workaround

Keep the previous `Menu` alive (e.g. via `menu.clone()` into a static) so the `Rc<RefCell<MenuChild>>` ref count stays > 0 until the next update cycle.

## Suggested fix

Replace `Cell<*const MenuChild>` with `Rc<MenuChild>` (as the FIXME already suggests). This would make the `MenuChild` lifetime tied to the `NSMenuItem`, preventing the use-after-free entirely.

divanshu-go:
The MudaMenuItem ObjC class stored a raw `Cell<*const MenuChild>` ivar pointing into the Rust heap. If the owning `Menu` was dropped or replaced while the native NSMenuItem was still visible (e.g. Screenpipe rebuilding the tray menu while it was open), clicking any stale item would dereference freed memory — manifesting as a panic in `String::as_str` or `slice::from_raw_parts` inside `fire_menu_item_click`.

Replace the ivar with `RefCell<Option<Rc<RefCell<MenuChild>>>>`. Each `create_ns_item_for_*` function now receives a cloned `Rc` of the owning `MenuChild` and stores it on the NSMenuItem. AppKit retains the NSMenuItem, so the `Rc` keeps the `MenuChild` alive for exactly as long as the native item exists — no lifetime mismatch possible.

Fixes: https://github.com/tauri-apps/muda/issues/328


drewbluewasabi:
Any updates on getting this merged? This is a majorly needed fix and has a small footprint.
@amrbashir, not sure if mentioning you will help but can we get a review on this?

src/platform_impl/macos/mod.rs
```diff
@@ -1026,8 +1030,7 @@ define_class!(
     #[unsafe(super(NSMenuItem))]
     #[name = "MudaMenuItem"]
     #[thread_kind = MainThreadOnly]
-    // FIXME: Use `Rc` or something else to access the MenuChild.
-    #[ivars = Cell<*const MenuChild>]
+    #[ivars = RefCell<Option<Rc<RefCell<MenuChild>>>>]
```
amrbashir: The outer most wrapper should still be `Cell`, `Cell<Option<...>>`

amrbashir (commented):
Thank you, could you also add a change file in `.changes` directory?

PR #363, opened 2026-06-17
fix(windows): dangling pointer when `remove_for_hwnd` called before drop

### LINKED ISSUE #15523: [bug] [windows] Use-after-free crash (`0xc000041d`) when `Window::set_menu` is called from a non-main thread — `unsafe impl Send + Sync` on menu types leaves the main-thread contract unenforced
### Describe the bug

On Windows, calling `Window::set_menu` (or rebuilding a menu) from a **non-main thread** — e.g. from an `AppHandle::listen` event handler, which runs on a tokio worker thread when the event is emitted from the frontend — intermittently crashes the **host process** later with exit code `0xc000041d` (`STATUS_FATAL_USER_CALLBACK_EXCEPTION`).

The root cause chain (confirmed by symbolized WER minidumps, see below):

1. Tauri's menu wrapper types are declared `unsafe impl Send + Sync` with the safety comment *"We make sure it always runs on the main thread"* (`tauri-2.11.2/src/menu/mod.rs`, lines 77-82). However, this contract is **not enforced anywhere**, and the public API happily accepts calls from any thread.
2. The underlying `muda::Menu` is `Rc<RefCell<platform_impl::Menu>>` (`muda-0.19.2/src/menu.rs`, lines 12-15) — a **non-atomic** refcount.
3. Inside `Window::set_menu` (`tauri-2.11.2/src/window/mod.rs`, lines 1276-1313) and `Window::remove_menu` (lines 1321-1352), closures holding clones of the old/new `muda::Menu` are posted to the main thread via `run_on_main_thread`, while the **calling thread** concurrently performs its own clones/drops (menu stash bookkeeping, `menu_lock().replace(...)`, dropping the returned previous menu).
4. When the caller is not the main thread, the non-atomic `Rc` refcount is mutated from two threads concurrently → refcount race → **premature free** of the platform `Menu` while the Win32 subclass installed by muda still holds its raw address (`dwrefdata`, `muda-0.19.2/src/platform_impl/windows/mod.rs`, lines 1090-1096).
5. The next menu-bar redraw delivers `WM_UAHDRAWMENU` (`0x91`) to `muda::menu_subclass_proc` (same file, lines 1148-1162), which dereferences the freed `Menu` and reads its `hwnds: RefCell<HashMap<isize, MenuTheme>>` → wild pointer read inside hashbrown's SSE2 group probe → access violation inside a window-procedure callback → the kernel terminates the process with `STATUS_FATAL_USER_CALLBACK_EXCEPTION`.

Because the failure is a refcount race, the crash is **intermittent** and **decoupled in time** from the `set_menu` call (it fires on a later `WM_NCACTIVATE`/menu redraw), which makes it very hard to attribute without a minidump. In our app it took four crash dumps over a month and a full dump-symbolization pass to localize; we initially suspected completely unrelated subsystems.

### Reproduction

Minimal shape (we can extract a standalone repro repository on request):

```rust
// setup hook (main thread) — initial menu, works fine
let menu = MenuBuilder::new(app).items(/* ... */).build()?;
main_window.set_menu(menu)?;

// the footgun: rebuild the menu from an event listener.
// Listeners for events emitted from the frontend run on a tokio worker thread.
app.handle().listen("locale-changed", move |_| {
    // ⚠ this closure is NOT on the main thread
    let menu = build_menu_again(&app_handle); // creates muda Rc objects off-main
    if let Some(window) = app_handle.get_webview_window("main") {
        let _ = window.set_menu(menu); // Rc clones/drops race with main-thread posts
    }
});
```

1. Frontend `emit('locale-changed')` (or any IPC-triggered event) several times — each triggers a menu rebuild on a worker thread.
2. Keep using the app; switch window focus / activate other windows (this forces menu-bar redraws via `WM_UAHDRAWMENU`).
3. Intermittently (for us roughly once per several rebuild-heavy sessions) the process dies with exit code `0xc000041d` minutes after the last rebuild.

**Workaround (confirmed):** wrapping the entire rebuild (menu construction + `set_menu`) in `app_handle.run_on_main_thread(...)` eliminates the crash. After this one-line-shape change, our previously-reproducing stress scenario (rapid theme/locale switches + file-open driven rebuilds + window activation cycles) produced zero crashes and zero new WER dumps.

### Expected behavior

One of (in descending order of preference):

1. `Window::set_menu` / `AppHandle::set_menu` / menu builders marshal **all** ownership operations of the underlying `Rc`-backed muda objects to the main thread (or muda switches to `Arc` + atomic refcounts), so that the existing `unsafe impl Send + Sync` is actually sound; or
2. Menu APIs detect off-main-thread usage and fail fast (`debug_assert!` / explicit `Error::MustRunOnMainThread`) instead of silently corrupting the refcount; or
3. At minimum, the documentation for `Menu`, `MenuBuilder`, `Window::set_menu`, and `AppHandle::listen` warns loudly that menu objects must never be created, cloned, dropped, or set outside the main thread, and that event listeners do not run on the main thread.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26200 x86_64 (X64)
    ✔ WebView2: 149.0.4022.62
    ✔ MSVC: Visual Studio Build Tools 2026
    ✔ rustc: 1.95.0 (59807616e 2026-04-14)
    ✔ cargo: 1.95.0 (f2d3ce0bd 2026-03-21)
    ✔ rustup: 1.29.0 (28d1352db 2026-03-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc
    - node: 24.14.1
    - npm: 11.11.0

[-] Packages
    - tauri 🦀: 2.11.2
    - tauri-build 🦀: 2.6.2
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.3
    - @tauri-apps/api ⱼₛ: 2.11.0
    - @tauri-apps/cli ⱼₛ: 2.11.2
    - muda (transitive): 0.19.2
```

### Stack trace

```text
Symbolized stack of the crashing (main) thread from a WER minidump of a dev build (`minidump-stackwalk` + app PDB). We have **four** dumps collected over a month — all with the identical signature (main thread, `EXCEPTION_ACCESS_VIOLATION_READ` at a non-canonical garbage address, faulting instruction `mov r8, qword [rdx]` = hashbrown SSE2 group load, message `0x91` = `WM_UAHDRAWMENU` in a register):


Crash reason:  EXCEPTION_ACCESS_VIOLATION_READ
Crash address: 0xffffffffffffffff (non-canonical: 0x7359ab83c8856573)
Crashing instruction: `mov r8, qword [rdx]`

Thread 0 main (crashed)
 0  app.exe!core::core_arch::x86::sse2::_mm_loadu_si128 [sse2.rs:1335]
 1  app.exe!hashbrown::raw::RawTable<tuple$<isize,muda::menu::MenuTheme>,...>::find [mod.rs:1236]
 2  app.exe!hashbrown::map::HashMap<isize,muda::menu::MenuTheme,...>::get [map.rs:1311]
 3  app.exe!std::collections::hash::map::HashMap<isize,muda::menu::MenuTheme,...>::get [map.rs:1002]
 4  app.exe!muda::platform_impl::platform::menu_subclass_proc [mod.rs:1153]
 5  comctl32.dll + 0x2ced7
 6  user32.dll + 0x3c918
 ...  (window procedure dispatch chain: user32 / comctl32 / uxtheme)
```

### Additional context

- The dangerous pattern is extremely easy to fall into: the natural way to implement "rebuild the native menu when the frontend changes locale/recent-files" is exactly an `AppHandle::listen` handler calling `set_menu`, and nothing in the type system or docs pushes back.
- The Win32 subclass registered by muda stores the raw address of the platform `Menu` as `dwrefdata` (`dwrefdata_from_obj`), so any premature free of the `Rc` allocation turns every subsequent `WM_UAHDRAWMENU`/`WM_NCACTIVATE` into a UAF. This is what makes the crash fire long after the actual race.
- Happy to provide the full minidumps privately, file an accompanying issue on `tauri-apps/muda` about the `Rc`/`dwrefdata` design, or PR a `debug_assert!`-style guard if maintainers indicate a preferred direction.

Legend-Master:
Fix https://github.com/tauri-apps/tauri/issues/15523

Spent so much time tracking this one down, the problem in https://github.com/tauri-apps/tauri/issues/15523 was not that the menu getting accessed off main thread but simply the drop orders

A simpler reproduction tauri code is just

```rust
{
    let _ = window.remove_menu().unwrap();
}
{
    let _ = window
        .set_menu(tauri::menu::Menu::default(&app_handle).unwrap())
        .unwrap();
}
```

The problem here is that we didn't clean up the window subclass when removing the menu (the drop didn't clean it up because the tracking hwnd is removed by `remove_for_hwnd`) and also we didn't refresh it in time that we called `SetMenu` before `SetWindowSubclass` in `init_for_hwnd_with_theme`, and `SetMenu` triggered `menu_subclass_proc` which contains the stale `Menu` pointer

---

sorry about the push to dev branch by accident

### DIFF
--- a/.changes/fix-windows-dangling-pointer.md
+++ b/.changes/fix-windows-dangling-pointer.md
@@ -1,0 +1,5 @@
+---
+"muda": patch
+---
+
+On Windows, fixed a bug that would truncate menubar items text to ~10 characters in dark-mode.
--- a/src/platform_impl/windows/mod.rs
+++ b/src/platform_impl/windows/mod.rs
@@ -112,9 +112,8 @@ pub(crate) struct Menu {
 
 impl Drop for Menu {
     fn drop(&mut self) {
-        let hwnds = self.hwnds.borrow().keys().copied().collect::<Vec<_>>();
-        for hwnd in &hwnds {
-            let _ = unsafe { self.remove_for_hwnd(*hwnd) };
+        for (hwnd, _) in self.hwnds.borrow_mut().drain() {
+            unsafe { Self::remove_from_hwnd(hwnd) };
         }
 
         fn remove_from_children_stores(internal_id: u32, children: &Vec<Rc<RefCell<MenuChild>>>) {
@@ -143,10 +142,6 @@ impl Drop for Menu {
         }
 
         unsafe {
-            for hwnd in &hwnds {
-                SetMenu(*hwnd as _, std::ptr::null_mut());
-                RemoveWindowSubclass(*hwnd as _, Some(menu_subclass_proc), MENU_SUBCLASS_ID);
-            }
             DestroyMenu(self.hmenu);
             DestroyMenu(self.hpopupmenu);
         }
@@ -346,13 +341,13 @@ impl Menu {
         self.hwnds.borrow_mut().insert(hwnd, theme);
 
         // SAFETY: HWND validity is upheld by caller
-        SetMenu(hwnd as _, self.hmenu);
         SetWindowSubclass(
             hwnd as _,
             Some(menu_subclass_proc),
             MENU_SUBCLASS_ID,
             dwrefdata_from_obj(self),
         );
+        SetMenu(hwnd as _, self.hmenu);
         DrawMenuBar(hwnd as _);
 
         Ok(())
@@ -367,12 +362,16 @@ impl Menu {
             .borrow_mut()
             .remove(&hwnd)
             .ok_or(crate::Error::NotInitialized)?;
-
-        // SAFETY: HWND validity is upheld by caller
-        SetMenu(hwnd as _, std::ptr::null_mut());
-        DrawMenuBar(hwnd as _);
-
+        Self::remove_from_hwnd(hwnd);
         Ok(())
+    }
+
+    unsafe fn remove_from_hwnd(hwnd: isize) {
+        let hwnd = hwnd as windows_sys::Win32::Foundation::HWND;
+        // SAFETY: HWND validity is upheld by caller
+        RemoveWindowSubclass(hwnd, Some(menu_subclass_proc), MENU_SUBCLASS_ID);
+        SetMenu(hwnd, std::ptr::null_mut());
+        DrawMenuBar(hwnd);
     }
 
     pub unsafe fn attach_menu_subclass_for_hwnd(&self, hwnd: isize) {


FabianLars (approved):

.changes/fix-windows-dangling-pointer.md
```diff
@@ -0,0 +1,5 @@
+---
+"muda": patch
+---
+
+On Windows, fixed a bug that would truncate menubar items text to ~10 characters in dark-mode.
```
Legend-Master, line 5: I swear I wrote the change file and it didn't look like this 🤦‍♂️

PR #341, opened 2026-04-17
Feat/gtk4

rmakestrash-jpg:
Gtk4

PR #369, opened 2026-07-03
feat(gtk): port the backend to gtk4

johncarmack1984:
Toward #270: the gtk4 port, targeting feat/gtk4 (#272) so the diff is what sits on top of that branch, in three commits: the port (@rmakestrash-jpg's continuation from #341 adapted to current dev), the icon rendering fix, and CI. Part of the delta is re-syncing shared modules with dev's current API (the KeyAccelerator line), which moved after the branch split.

The API: GTK4 deleted `gtk::Menu`, `gtk::MenuBar`, and `gtk::Container`, so those types can't survive. Every method keeps its name and shape. `gtk_context_menu()` returns `gtk::PopoverMenu`, `gtk_menubar_for_gtk_window` returns `Option<gtk::PopoverMenuBar>`, `Container` bounds widen to `Widget`. gtk3 call sites compile unchanged -- checked against tauri and tauri-apps/tao#1258 -- and no gtk3 process can link a gtk4 muda anyway.

Icons: `PopoverMenu` ignores icons set on model items, so icon items get a `custom` slot with a real icon+label widget bound via `add_child`. Here it is in a running Tauri app on the full gtk4 stack (tao#1258 + wry gtk4 + this); label, icon, and checked changes show up live:

menu icon rendering in a Tauri app on the gtk4 stack

Also fixes the macOS break from #341 (`KeyAccelerator` throughout) and drops the ksni and sync-workflow changes per the feedback there. check, clippy, and tests green on Linux (GTK 4.18) and macOS. MSRV 1.92; CI needs `libgtk-4-dev`. The two changed return types are the main review question.


johncarmack1984:
CI should be in better shape now. Three things were failing: the test jobs pin a 1.73 toolchain that can't parse a v4 lockfile (MSRV is 1.92 with this port, so the pin moves with it), the runners were missing libgtk-4-dev, and the clippy gate caught one lint plus examples still calling the gtk API with gtk3-era tao types; those calls are commented until a tao release on gtk4 exists. Everything green locally on the exact CI commands for Linux and macOS.

The audit failure is pre-existing rather than from this PR: quick-xml < 0.41 arrives through winit -> wayland-scanner in the dev-dependencies, the advisory is dated June 29, and dev's lockfile carries the same version, so a fresh audit fails identically on dev. Nothing here can reach the patched version until wayland-scanner bumps.

amrbashir:
can you target my `feat/gtk4` branch? This way I can review and discuss what you added on top of my work without the noise of previous work.

johncarmack1984:
Done. The PR now targets feat/gtk4, with the delta in three commits on top of your branch: the port (rmakestrash-jpg's continuation adapted to current dev), the icon rendering fix, and CI. One note on the diff: dev's shared modules moved to the KeyAccelerator API after the branch split, so part of the delta is re-syncing those files with current dev.

johncarmack1984:
The audit failure predates the branch: RUSTSEC-2026-0194 and RUSTSEC-2026-0195, quick-xml coming in through wayland-scanner in the winit dev-deps. dev resolves the same quick-xml version, so it fails the same way. It heals here with no changes once Smithay/wayland-rs#938 ships in a wayland-scanner release. The rest of the matrix is green.

amrbashir:
Thanks for your PR, but after a look at your changes, I decided to just continue implementing remaining parts in https://github.com/tauri-apps/muda/pull/272 myself as it will take a lot of time to review and go back and forth just to get the code to a state where I can accept it and maintain going forwards and I would like to finish gtk4 migration across the ecosystem as fast as I can.

Apologies for this inconvenience.

PR #380, opened 2026-07-28
feat!: support native icons on Windows and GTK backends

amrbashir:


examples/wry.rs
```diff
@@ -123,6 +123,14 @@ fn main() -> wry::Result<()> {
...
         Some(Accelerator::new(Some(Modifiers::CONTROL), Code::KeyC)),
     );
+    #[cfg(target_os = "macos")]
+    let native_icon = "NSFolder".to_string();
+    #[cfg(target_os = "windows")]
+    let native_icon = "SIID_FOLDER".to_string();
+    #[cfg(not(any(target_os = "macos", target_os = "windows")))]
+    let native_icon = "folder-symbolic".to_string();
+    let native_icon_item =
+        IconMenuItem::with_native_icon("Native icon", true, Some(native_icon), None);
```
Legend-Master, line 133: I get the reason for this but I feel like this is quite hard on the developers, could we do some abstraction with our old enum with platform specific variants? I don't like how we map the strings in this PR.

amrbashir: Can you explain what you mean by the mapping here? the examples or the match arm on Windows backend ?

I don't like an enum here because linux theme icons could be virtually anything, macOS and Windows also don't map one to one which why I landed on a string based solution.

Another solution to consider is 
```rs

struct NativeIcon(PlatformNativeIcon);

impl NativeIcon {
  fn from_id(id: u32); // Windows
  fn from_name(name: &str); //macOS and Linux
}
```

This would remove the match arm on Windows backend.

amrbashir: I get that this is bad for DX, so maybe we could have both string and enum based solutions. Will open another PR with some ideas later.

Legend-Master: > Can you explain what you mean by the mapping here? the examples or the match arm on Windows backend ?

Both, you need to set the icon for each platform differently and the string match arm on Windows.

> I get that this is bad for DX, so maybe we could have both string and enum based solutions. Will open another PR with some ideas later.

Yeah, that sounds great, I was thinking something like:

```rust
enum NativeIcon {
   Add,
   Computer,
   /// Any string
   Custom(String)
}
```

amrbashir: https://github.com/tauri-apps/muda/pull/382

PR #382, opened 2026-07-29
feat: bring back NativeIcon enum while keeping support on Windows and Linux

amrbashir:


src/icon.rs
```diff
@@ -164,3 +164,171 @@ impl Icon {
...
+    Raw(windows_sys::Win32::UI::Shell::SHSTOCKICONID),
+    /// A platform-specific native icon name.
+    #[cfg(not(windows))]
+    Raw(String),
+}
+
+impl NativeIcon {
+    /// Creates a native icon from a Windows `SHSTOCKICONID`.
+    #[cfg(windows)]
+    pub fn from_id(id: windows_sys::Win32::UI::Shell::SHSTOCKICONID) -> Self {
```
Legend-Master, line 304: will it be a bit more clear to name it to something like `from_sh_stock_icon_id`? since it's gated behind `#[cfg(windows)]`, `from_id` also works for me.

Legend-Master (commented):
awesome

src/icon.rs
amrbashir: I think from_id is enough, from_sh_stock_icon_id is too verbose for no reason

PR #389, opened 2026-08-08
feat: make MenuEvent::send public

0-don:
A menu can be rendered by something other than muda. A tray serving its menu over StatusNotifierItem builds the items from muda's model but receives the clicks itself, and currently has no way to report them back: `send` is the only path to both the handler and the receiver channel, and it is `pub(crate)`.

Needed by [tray-icon#341](https://github.com/tauri-apps/tray-icon/pull/341), which adds a StatusNotifierItem backend so left clicks work on Linux. Without this the backend can serve the menu but cannot deliver activation events, so `MenuEvent::receiver()` stays silent.

No behaviour change; only the visibility of an existing function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -587,7 +587,12 @@ impl MenuEvent {
         }
     }
 
-    pub(crate) fn send(event: MenuEvent) {
+    /// Dispatches a menu event to the handler, or to the channel returned by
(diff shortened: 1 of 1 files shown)


amrbashir:
this is no longer needed, I have went with a different approach and ksni backend is now added to tray-icon and will be released soon

PR #404, opened 2026-09-09
refactor: make gtk3 and gtk4 exclusivity error to just a warning with fallback to noop backend

amrbashir:


FabianLars:
where does this idea come from? do other crates do this? looks so weird to me ngl. isn't this something you _want_ to hard fail on so it does not semi-silently succeeds to build and then not work at runtime? especially on linux which many devs only build on ci

amrbashir:
In tauri workspace, we have tauri-runtime-wry using gtk3 and tauri-runtime-cef using gtk4, because Cargo unifies the dependencies, and unifies the features, so muda ends up with both features enabled which makes `cargo build` and `cargo test` fails in tauri repo.

Same thing for tray-icon but that one is being fixed in https://github.com/tauri-apps/tray-icon/pull/364

In hindsight, this is a really bad decision, as it is is easy to use muda default features and enable gtk4 and never notice the warning. I will find a better solution or revert.

amrbashir:
Same technique as tray-icon can be applied to muda, and we can just use gtk4 as fallback when both are enabled. #405

